PR #65, opened 2020-01-26
Removed and ignored local.yml to make it truly local

georgehelman:
This was kind of annoying me. I wanted to bind a port for the Postgres database but didn't want to force that on everybody else too. I suggest everyone manage their own local compose file and new developers can copy from the default docker-compose.yml


copelco:
Without an example, I think it would be hard to build ``local.yml`` manually. We may just want to rename it to ``local.yml.example`` so people can copy it.

copelco:
I think is resolved with the updated ``docker-compose.yml``

PR #66, opened 2020-01-28
Sendgrid

georgehelman:


georgehelman:
It says the api key env variable isn't set but I put it in the env file. What do I need to do to make Travis love me?

PR #67, opened 2020-02-09
Track upload user

copelco:
* add ``Batch.user``
* add ``Batch.date_uploaded`` (with migrations to update existing records)
* add ``Batch.label`` (with migrations to update existing records)
* add Batch admin with CIPRSRecord inline


dear_petition/petition/forms.py
```diff
@@ -12,17 +12,21 @@
...
             record.data = record.parse_report(file_)
             if "error" in record.data:
                 raise forms.ValidationError(record.data["error"])
             if "Defendant" in record.data and "Name" in record.data["Defendant"]:
-                record.label = record.data["Defendant"]["Name"]
+                label = record.data["Defendant"]["Name"]
+                record.label = label
+                if idx == 0:
+                    batch.label = label
+                    batch.save()
```
georgehelman: I think this should be wrapped in a transaction to prevent the DB from being left in a weird state in case of failure.

copelco: Good idea

Issue #72, opened 2020-02-16
Implement permissions


georgehelman:
Right now, technically anyone can see records used to generate petitions in the tool. This was a concern that was raised during the meeting with Gina. Either we need to partition user files/cases so each user can only see their own or only admins can see the records. We will need to discuss the approach we want to take next meeting.

PR #75, opened 2020-02-19
Permissions

georgehelman:
#72

Still need to implement tests.


dear_petition/petition/views.py
```diff
@@ -23,6 +25,8 @@ def upload_report(request):
 @login_required
 def view_record(request, pk):
     record = get_object_or_404(CIPRSRecord, pk=pk)
+    if not is_owner(request.user,record.batch) and not request.user.is_superuser:
```
copelco: Only suggestion is to move ``request.user.is_superuser`` into the ``is_owner`` function for DRY, e.g. ``if not is_owner(request, batch)``

georgehelman:
Wrote a test and also added some factories and test utilities that should be useful in the future.

PR #84, opened 2020-03-23
Fixed bug where different parsers represent offense datetime differently

georgehelman:
500 error was happening because the most recent record compares dates by stripping the datetime, but it was chugging because some data in that field is represented as datetime, while others are represented as just dates, depending on which parser it came from. I prefer this change rather than modifying the most_recent_record function because I think it will make it easier to code in the future if we keep our data formats consistent.


georgehelman:
Wrong change, wrong project. Closing.

PR #92, opened 2020-04-16
[TSC 100] bootstrap frontend

michael-caktus:
Adds frontend/ directory full of a react app.

React app bootstrapped with CRA.

Adds initial dependencies including:
1. [React-Router](https://reacttraining.com/react-router/web/guides/quick-start) for routing
2. [styled-components](https://styled-components.com/) for styling
3. [Framer Motion](https://www.framer.com/api/motion/) for animations
4. some react testing-library stuff


frontend/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# pull official base image
+FROM node:13.12.0-alpine
+
+# set working directory
+WORKDIR /app
+
+# add `/app/node_modules/.bin` to $PATH
+# ENV PATH /app/node_modules/.bin:$PATH
+
```
Audiosutras: Are these lines no longer needed?

michael-caktus:
These last two commits add the frontend to local.yml such that it will start with all the other services. You can still start the frontend from the frontend directory. There's a note in frontend/README.md about adding setup instructions to point the local frotnend to staging API. We don't have that setup yet, so it's a TODO.

As for serving the app at /, it looks like we've moved away from that approach.

frontend/src/components/pages/HomePage/HomePage.js
```diff
@@ -0,0 +1,94 @@
...
+
+  const handleRemoveFile = rmFile => {
+    // browser stores a "path" to the last file uploaded on the input.
+    // It's necessary to "clear" the inputs value here, but not on drop--
+    // the browser just replaces previous files in the case of a drop.
+    if (fileInputRef.current) fileInputRef.current.value = '';
+    // TODO: It would be great to get this taken care of inside DragNDrop,
+    // TODO: but currently DragNDrop has no concept of a FilesList or removing files.
+    setFiles(files.filter(fl => fl.name !== rmFile.name));
+  };
```
Audiosutras, line 43: If I upload the same file multiple times, when I click to delete the duplicate(s) all files with that matching name are deleted. Only the file whose delete button has been clicked on should be deleted.

michael-caktus: Good call. This commit below 👇uses a happy marriage between Sets and the File object to enforce both the order of the list of files (so we can safely delete by index) and the uniqueness of files uploaded. Two birds with one stone!

PR #93, opened 2020-04-16
[TSC-102] frontend layout and routing

michael-caktus:
**Depends on TSC 100. Merge that first.** 

Initial project structure and basic layout and routing.

Only need two routes.
1. '/' which renders "Home if authenticated or "Login" if not.
-- "Home" will show the CIRPS uploader

2. `/generate`, which will contain the bi-directional multi-stage form view for generating a petition.
-- eventually `/generate/:batchId` will be a thing. Details will come with implementation.


frontend/public/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
```
Audiosutras, line 1: Can we get Django to serve this page?

frontend/src/components/App.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { AppStyled } from './App.styled';
+import GlobalStyle from '../styles/GlobalStyle';
+
+// Routing
+import { BrowserRouter, Switch } from 'react-router-dom';
+import ProtectedRoute from './containers/ProtectedRoute';
```
Audiosutras, line 7: I really like this concept of a `ProtectedRoute`. I'm going to start using it.

frontend/src/components/App.js
```diff
@@ -0,0 +1,33 @@
...
+
+// Pages
+import HomePage from './pages/HomePage/HomePage';
+import GenerationPage from './pages/GenerationPage/GenerationPage';
```
Audiosutras, line 11: `GeneratePDFPage`

frontend/src/components/containers/README.md
```diff
@@ -0,0 +1,5 @@
+### Containers
```
Audiosutras, line 1: Is calling your logical components `containers` a convention? Inside of the `components` directory, why not have to directories called `logical` and `presentational`?

frontend/src/components/elements/Button/Button.js
```diff
@@ -0,0 +1,26 @@
...
+
+  const handleMouseUp = e => onClick(e);
+
+  return <ButtonStyled {...props} onMouseUp={handleMouseUp} >{children}</ButtonStyled>;
+}
+
+/* Props */
+export const POSITIVE = "positive";
+export const CAUTION = "caution";
+export const NEUTRAL = "neutral";
```
Audiosutras, line 15: Can we create a `constants.js` file that contains all of our constants on the frontend. We eventually may also want to have `utils.js` file that has all of our utility functions.

Audiosutras: So can we import `POSITIVE`, `CAUTION`, and `NEUTRAL` from that file.

frontend/src/components/elements/Button/Button.styled.js
```diff
@@ -0,0 +1,62 @@
...
+const caution = css`
+  background: ${colorCaution};
+  border: 1px solid ${colorCaution};
+  color: ${colorWhite};
+`;
+
+const neutral = css`
+  background: ${colorWhite};
+  border: 1px solid ${colorPrimary};
+  color: ${colorPrimary};
```
Audiosutras, line 61: Perhaps we can have `utils` directory, I think this could be included in our styled-component utility functions.

frontend/src/components/containers/README.md
Audiosutras: Mainly just asking the question, not requesting a change here

frontend/src/components/elements/Button/Button.js
michael-caktus: I think we'll probably end up with a constants file, or several of them. But I like the idea of keeping constants scoped. The nice thing about the way this is done here is you can do this:
`import Button, { POSITIVE, CAUTION, NEUTRAL } from "./Button/Button"`

It makes it clear to the user what those states are for and how they should be used.

frontend/src/components/containers/README.md
michael-caktus: "containers" is the name by convention. It's a bit less specific than "logical", which might tend to make it a dumping ground for any component with logic.  They are logical components, but their main distinguishing feature is that they do not render display-level JSX, they render other components. Essentially, a container is something that "Contains" presentational components. Maybe sounds like splitting hairs, but I think I can see why the convention is to call them "Containers".

PR #99, opened 2020-04-17
Upgrade #7

copelco:


config/settings/base.py
```diff
@@ -241,7 +241,7 @@
 # https://django-allauth.readthedocs.io/en/latest/configuration.html
 ACCOUNT_EMAIL_REQUIRED = True
 # https://django-allauth.readthedocs.io/en/latest/configuration.html
-ACCOUNT_EMAIL_VERIFICATION = "mandatory"
+ACCOUNT_EMAIL_VERIFICATION = "optional"
```
Audiosutras, line 244: Instead I was thinking we should add IS_REVIEW environment variable. If it is true we then can change `ACCOUNT_EMAIL_VERIFICATION` to "optional" otherwise in production it should be "mandatory"

PR #103, opened 2020-04-25
[DP-20] Fix prod migrate

copelco:
Suggested changes to get migrations to run against a production database dump.

### DIFF
--- a/dear_petition/petition/utils.py
+++ b/dear_petition/petition/utils.py
@@ -38,8 +38,7 @@ def make_datetime_aware(dt_str):
     if dt_str is None or dt_str == "":
         return None
     # a) Turn the datetime string into a timezone-naive datetime object
-    for fmt in DATETIME_FORMAT:
-        naive_datetime_obj = dateutil.parser.parse(dt_str)
+    naive_datetime_obj = dateutil.parser.parse(dt_str)
 
     # b) Assign the timezone-naive datetime a timezone based on settings.TIME_ZONE
     aware_datetime_obj = make_aware(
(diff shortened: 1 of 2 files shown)


dear_petition/petition/migrations/0021_merge_20200425_1737.py
```diff
@@ -0,0 +1,13 @@
+# Generated by Django 2.2.10 on 2020-04-25 17:37
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("petition", "0020_comment_time"),
+        ("petition", "0020_offense_offenserecord"),
```
Audiosutras, line 10: I think we should cherry-pick this `commit cf993dcc69318d09082458bd8bdb0e3bab8d09a2` from the upgrade branch. It re-orders the comment migrations putting them after the offense/offenserecord migration (21-23)

dear_petition/petition/utils.py
```diff
@@ -36,7 +38,9 @@ def make_datetime_aware(dt_str):
     if dt_str is None or dt_str == "":
         return None
     # a) Turn the datetime string into a timezone-naive datetime object
-    naive_datetime_obj = datetime.strptime(dt_str, DATETIME_FORMAT)
+    for fmt in DATETIME_FORMAT:
+        naive_datetime_obj = dateutil.parser.parse(dt_str)
+
```
Audiosutras, line 43: Why do we need to iterate through the `DATETIME_FORMAT` string in order to come up with the `naive_datetime_obj`? Can we remove line 41?

copelco: Yes, good catch

PR #104, opened 2020-04-27
DP-21 

Audiosutras:
This PR adds code to `create` `Offense` & `OffenseRecord` instances when `CIPRSRecord.refresh_record_from_data` is called and removes similar logic from `UploadFileForm`.

### DIFF
--- a/dear_petition/petition/models.py
+++ b/dear_petition/petition/models.py
@@ -168,12 +168,15 @@ class CIPRSRecord(models.Model):
                 disposition_method=offenses.get("Disposition Method", ""),
             )
             offense_records = offenses.get("Records", [])
+            if not created:
+                existing_offense_records = OffenseRecord.objects.filter(offense=offense)
+                existing_offense_records.delete()
             for offense_record in offense_records:
                 try:
                     code = int(offense_record.get("Code"))
                 except ValueError:
                     code = None
-                o_record, created = OffenseRecord.objects.get_or_create(
+                OffenseRecord.objects.create(
                     offense=offense,
                     law=offense_record.get("Law", ""),
                     code=code,
--- a/dear_petition/petition/models.py
+++ b/dear_petition/petition/models.py
@@ -54,7 +54,6 @@ class CIPRSRecordManager(models.Manager):
         """
         ciprs_record = CIPRSRecord(batch=batch, label=label, data=data)
         ciprs_record.refresh_record_from_data()
-        ciprs_record.save()
 
 
 class CIPRSRecord(models.Model):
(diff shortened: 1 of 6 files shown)


dear_petition/petition/models.py
```diff
@@ -168,12 +168,15 @@ def refresh_record_from_data(self):
                 disposition_method=offenses.get("Disposition Method", ""),
             )
             offense_records = offenses.get("Records", [])
+            if not created:
```
dchukhin, line 171: We should probably call this regardless of the value of `created`, because whether we're creating a new `Offense` here or not, we don't want other `OffenseRecord`s for this `Offense` hanging out in the database.

dear_petition/petition/models.py
```diff
@@ -168,12 +168,15 @@ def refresh_record_from_data(self):
                 disposition_method=offenses.get("Disposition Method", ""),
             )
```
dchukhin, line 169: If there are more `Offense`s for this `CIPRSRecord` in the database, should we delete them here too?

dear_petition/petition/tests/test_models.py
```diff
@@ -222,20 +243,75 @@ def test_refresh_record_from_data():
...
+    ciprs_record.data.get("Offense Record").update(updated_o_records)
+    ciprs_record.refresh_record_from_data()
+
+    raw_data_offense_records = ciprs_record.data.get("Offense Record").get("Records")
+    stored_codes = []
+    stored_offense_records = list(ciprs_record.offenses.first().offense_records.all())
+    for offense_record in stored_offense_records:
+        stored_codes.append(offense_record.code)
+    assert len(stored_offense_records) == len(raw_data_offense_records)
+    assert len(stored_actions) == len(raw_data_actions)
```
dchukhin, line 312: This line is supposed to be `assert len(stored_codes) == len(raw_data_codes)`, right?

Audiosutras: I actually think I forgot to remove that line of the code. I just want to have the assertions above 312 and below it.

dear_petition/petition/models.py
Audiosutras: done

dear_petition/petition/models.py
Audiosutras: yep

Audiosutras: done

dear_petition/petition/models.py
```diff
@@ -161,16 +161,22 @@ def refresh_record_from_data(self):
         # make changes to a users a offenses (I doubt it) then we need
         # think about revising the methods chosen here.
         offenses = self.data.get("Offense Record", {})
+
+        # Check and remove any existing Offense Records and the existing Offenses
+        # before reading from raw_data
+        existing_offense_records = OffenseRecord.objects.filter(
+            offense=self.offenses.first()
```
dchukhin, line 168: Since there can be multiple `Offense`s for a `CIPRSRecord`, we should make sure to handle all of them here (something like `offense__ciprs_record=self)`

dear_petition/petition/tests/test_models.py
```diff
@@ -148,6 +213,106 @@ def test_refresh_record_from_data():
...
+    }
+    ciprs_record.data.get("Offense Record").update(updated_o_records)
+    ciprs_record.refresh_record_from_data()
+
+    raw_data_offense_records = ciprs_record.data.get("Offense Record").get("Records")
+    stored_codes = []
+    stored_offense_records = list(ciprs_record.offenses.first().offense_records.all())
+    for offense_record in stored_offense_records:
+        stored_codes.append(offense_record.code)
+    assert len(stored_offense_records) == len(raw_data_offense_records)
```
dchukhin, line 311: Could you add  an `assert len(stored_codes) == len(o_record_int_code_list)` here, just to make sure that `stored_codes` and `o_record_int_code_list` really have the same elements?

Audiosutras: sure

Audiosutras: done

dear_petition/petition/models.py
Audiosutras: how interesting. I didn't know you could do that. done

PR #110, opened 2020-05-20
[DP 34/map petitioner] 

Audiosutras:
This PR further refactors what was `data_dict.py`. This PR takes code related to the petitioner and puts it in the `map_petitioner` definition. Note: age is commented out in map_petitioner and the associated `AGE` test is commented out. Once we store `age` as field on the CIPRSRecord we should uncomment these lines.

https://caktus.atlassian.net/browse/DP-34

### DIFF
--- a/dear_petition/petition/export/tests/test_mapper.py
+++ b/dear_petition/petition/export/tests/test_mapper.py
@@ -2,11 +2,17 @@ import pytest
 
 from dear_petition.petition import constants
 from dear_petition.petition.export import mapper
+from dear_petition.petition.utils import make_datetime_aware
+
+from dear_petition.petition.tests.factories import (
+    BatchFactory,
+    CIPRSRecordFactory,
+)
 
 
 pytestmark = pytest.mark.django_db
 
-
+# map_petition test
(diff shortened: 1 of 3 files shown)


dear_petition/petition/export/tests/test_mapper.py
```diff
@@ -29,3 +35,66 @@ def test_map_petition__district(petition):
     mapper.map_petition(data, petition)
     assert data["District"] == "Yes"
     assert data["Superior"] == ""
+
+
+# map_petitioner tests
+def test_map_petitioner__name(petition):
```
copelco, line 41: I think we can use the ``batch`` and ``record1``/``record2`` fixtures here

dear_petition/petition/export/tests/test_mapper.py
```diff
@@ -29,3 +35,66 @@ def test_map_petition__district(petition):
...
+    data = {}
+    batch = BatchFactory()
+    record = CIPRSRecordFactory(batch=batch)
+    record.refresh_record_from_data()
```
copelco, line 45: I wonder if we want to use ``refresh_record_from_data``, since it's static in the factory.

PR #111, opened 2020-05-20
[DP-30] Dismissed petition support

copelco:
* Add ``petition.types`` package to encapsulate business logic for each type of petition
* Add ``petition.types.dismissed`` module for AOC-CR-287
* Add ``create_batch_petitions`` to ETL to create generatable petitions
* Add model helper functions like ``batch.dismissed_offense_records`` and ``petition.get_offense_records``


dear_petition/petition/migrations/0025_auto_20200520_0120.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 2.2.12 on 2020-05-20 01:20
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
```
Audiosutras, line 9: We just need to be mindful that we don't have migration conflicts between this PR and DP-35

dear_petition/petition/models.py
```diff
@@ -210,6 +210,8 @@ class Offense(models.Model):
     )
     disposed_on = models.DateField(blank=True, null=True)
     disposition_method = models.CharField(max_length=256)
+    verdict = models.CharField(max_length=256, blank=True)
+    plea = models.CharField(max_length=256, blank=True)
```
Audiosutras, line 214: NOTE as a reminder for me while I review: Are we currently extracting `verdict` and `plea` from the JSONBlob yet? Can do we write to these fields?

PR #113, opened 2020-05-23
[DP-36]

Audiosutras:
This PR:

- adds body logic to `map_offenses`
- adds basic test for `map_offenses`
- removes no longer used code associated with getting offense records
- `confest.py` and `factories.py` has also been updated to better support testing `map_offenses`

This PR should conclude work in `mapper.py` for this phase (DP-34 - DP-36)


dear_petition/petition/constants.py
```diff
@@ -42,3 +42,16 @@
 )
 
 NORTH_CAROLINA = "NC"
+DURHAM_COUNTY = "DURHAM"
+
+DISTRICT_COURT_WITHOUT_DA_LEAVE = "Dismissal without Leave by DA"
+
+DISPOSITION_METHODS = (
```
copelco, line 49: If we're moving this to constants, I think we should prepend with dismissed:

```suggestion
DISMISSED_DISPOSITION_METHODS = (
```

dear_petition/petition/export/mapper.py
```diff
@@ -78,6 +78,32 @@ def map_agencies(data, petition, extra={}):
 
 
 def map_offenses(data, petition, extra={}):
-    pass
     offense_records = petition.get_offense_records()
-    print(offense_records)
+    for idx, offense_record in enumerate(offense_records, 1):
+        # The index of the offense determines what line on the petition form
+        # the offense will be on
+        formatted_arrest_date = (
```
copelco, line 85: Nit picky, but having a utility function here could help with readability here, something like:

```python
def petition_date(date):
    return date.strftime(constants.DATE_FORMAT) if date else ""

# then:
offense = offense_record.offense
formatted_arrest_date = petition_date(offense.ciprs_record.arrest_date)
formatted_offense_date = petition_date(offense.ciprs_record.offense_date)
formatted_disposed_on = petition_date(offense.disposed_on)
```

dear_petition/petition/constants.py
Audiosutras: done

dear_petition/petition/export/mapper.py
Audiosutras: done

PR #114, opened 2020-05-24
DP-28

Audiosutras:
Overrode `get_queryset` on `BatchViewSet` to restrict user read access to the Batches they have created. The exception is superusers who can view all batches.

### DIFF
--- a/dear_petition/petition/api/viewsets.py
+++ b/dear_petition/petition/api/viewsets.py
@@ -44,13 +44,21 @@ class ContactViewSet(viewsets.ModelViewSet):
 
 
 class BatchViewSet(viewsets.ModelViewSet):
-
     queryset = petition.Batch.objects.prefetch_related(
         "petitions", "records__offenses__offense_records"
     )
     serializer_class = serializers.BatchSerializer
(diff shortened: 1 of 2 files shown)


dear_petition/petition/api/viewsets.py
```diff
@@ -44,14 +44,22 @@ class ContactViewSet(viewsets.ModelViewSet):
...
         "petitions", "records__offenses__offense_records"
     )
     serializer_class = serializers.BatchSerializer
     permission_classes = [permissions.IsAuthenticated]
     parser_classes = (parsers.MultiPartParser, parsers.FormParser)
 
+    def get_queryset(self):
+        """ Filter queryset so that user's only have read access on objects they have created
+        """
+        if self.request.user.is_superuser:
```
copelco, line 57: I might flip this, something like:
```suggestion
        qs = super().get_queryset()
        if not self.request.user.is_superuser:
            qs = qs.filter(user=...)
        return qs
```

Audiosutras: done

Issue #119, opened 2020-05-26
Populate disposition method codes in petition Disposition column


copelco:
In export app, offense dispositions need to be mapped to the following codes within the petition PDF:

- Dismissal without Leave by DA (VD)
- Dismissed by Court (DC)
- Deferred Prosecution Dismissal (DPD)
- Discharge and Dismissal (DD)
- Conditional Discharge (DD)
- No Probable Cause (NPC)
- Never To Be Served (NTBS)
- Guilty to Lesser (Glty to Lesser)

PR #137, opened 2020-06-14
DP-119 Populate petition with disposition codes

georgehelman:
https://github.com/deardurham/dear-petition/issues/119


dear_petition/petition/constants.py
```diff
@@ -55,3 +57,16 @@
     "No Probable Cause",
     "Never To Be Served",
 )
+
+KNOWN_DISPOSITION_METHOD = {
```
copelco, line 61: The build is failing. I think this might be part of the reason?

```suggestion
KNOWN_DISPOSITION_METHODS = {
```

PR #149, opened 2020-06-29
Offense API returns plea and verdict

georgehelman:


dear_petition/petition/models.py
```diff
@@ -126,6 +126,8 @@ class OffenseRecord(models.Model):
     action = models.CharField(max_length=256)
     severity = models.CharField(max_length=256)
     description = models.CharField(max_length=256)
+    plea = models.CharField(max_length=256, null=True)
```
copelco, line 129:
```suggestion
    plea = models.CharField(max_length=256, blank=True)
```

Django suggests to [avoid using null on string-based fields such as CharField](https://docs.djangoproject.com/en/3.0/ref/models/fields/#null).

dear_petition/petition/models.py
```diff
@@ -126,6 +126,8 @@ class OffenseRecord(models.Model):
...
+    verdict = models.CharField(max_length=256, null=True)
```
copelco, line 130:
```suggestion
    verdict = models.CharField(max_length=256, blank=True)
```
Same as above

dear_petition/petition/etl/load.py
```diff
@@ -26,6 +26,7 @@ def import_ciprs_records(files, user):
             record.refresh_record_from_data()
             if settings.CIPRS_SAVE_PDF:
                 record.report_pdf = file_
+                record.save()
```
copelco, line 29: Good catch!

PR #155, opened 2020-07-12
Fix issue where we were displaying UTC datetimes on petition

georgehelman:


dear_petition/petition/api/viewsets.py
```diff
@@ -71,13 +71,6 @@ def get_queryset(self):
             qs = qs.filter(user=self.request.user)
         return qs
 
-    def create(self, request, *args, **kwargs):
```
copelco, line 74: Why are we removing this method?

dear_petition/petition/api/tests/test_batch.py
```diff
@@ -19,7 +19,6 @@ def test_batch_post_file(api_client, fake_pdf, mock_import):
     response = api_client.post(reverse("api:batch-list"), data=data)
     assert response.status_code == status.HTTP_201_CREATED
     assert mock_import.assert_called_once
-    assert "id" in response.data
```
copelco, line 22: These tests should remain

copelco (approved):

PR #161, opened 2020-08-07
Offense API returns plea and verdict

georgehelman:


dear_petition/petition/migrations/0030_auto_20200627_2339.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ("petition", "0029_offense_jurisdiction"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="offenserecord",
+            name="plea",
+            field=models.CharField(max_length=256, null=True),
```
copelco, line 16: These both say ``null=True``, but should be ``blank=True``

PR #170, opened 2020-09-27
Populate not guilty charges

georgehelman:


dear_petition/petition/etl/load.py
```diff
@@ -48,6 +48,15 @@ def create_batch_petitions(batch):
...
     # TODO: Not guilty
+    not_guilty_records = batch.not_guilty_offense_records()
+    petition_types = identify_distinct_petitions(not_guilty_records)
+    for petition_type in petition_types:
+        petition = batch.petitions.create(
+            form_type=NOT_GUILTY,
+            jurisdiction=petition_type["jurisdiction"],
+            county=petition_type["county"],
+        )
+        link_offense_records_and_attachments(petition)
```
robert-w-gries, line 59: Can you make a separate function that performs this logic? Something like

```
def create_petitions_from_records(records):
    petition_types = identify_distinct_petitions(records)
    for petition_type in petition_types:
        petition = batch.petitions.create(
            form_type=NOT_GUILTY,
            jurisdiction=petition_type["jurisdiction"],
            county=petition_type["county"],
        )
        link_offense_records_and_attachments(petition)

...
create_petitions_from_records(batch.dismissed_offense_records())
create_petitions_from_records(batch.not_guilty_offense_records())
```

dear_petition/petition/types/tests/test_not_guilty.py
```diff
@@ -0,0 +1,59 @@
...
+    OffenseRecordFactory,
+    PetitionFactory,
+)
+
+pytestmark = pytest.mark.django_db
+
+
+def test_charged_not_guilty_record(batch, record1):
+    """Charged not guilty records should be included"""
+    offense = OffenseFactory(verdict="Not Guilty", ciprs_record=record1)
```
robert-w-gries, line 16: Is there a difference between this `offense` and the test fixture `not_guilty_offense`?

requirements/base.txt
```diff
@@ -6,6 +6,7 @@ whitenoise==4.1.2  # https://github.com/evansd/whitenoise
 redis==3.2.1  # https://github.com/antirez/redis
 celery==4.3.0  # pyup: < 5.0  # https://github.com/celery/celery
 flower==0.9.3  # https://github.com/mher/flower
+vine==1.3.0
```
robert-w-gries, line 9: Was this meant to be added?

dear_petition/petition/models.py
```diff
@@ -123,8 +123,6 @@ class OffenseRecord(models.Model):
     action = models.CharField(max_length=256)
     severity = models.CharField(max_length=256)
     description = models.CharField(max_length=256)
-    plea = models.CharField(max_length=256, blank=True)
-    verdict = models.CharField(max_length=256, blank=True)
```
robert-w-gries, line 127: Was curious why this was being removed. Were these fields ever used? Did #163 remove the need for these fields?

robert-w-gries (changes requested):

dear_petition/petition/models.py
georgehelman: I removed these fields here because they were redundant with the plea and verdict on the Offense model. My current understanding is that plea and verdict is an offense-level field, so I removed them here but they are still on the Offense model.

requirements/base.txt
georgehelman: Yeah, our build was failing because pytest is looking for a module in vine that doesn't exist in the version it uses as a dependency, so I'm just pinning it to the version we want here.

dear_petition/petition/types/tests/test_not_guilty.py
georgehelman: No, good call. I'll use that.

dear_petition/petition/etl/load.py
georgehelman: Yep, good idea.

requirements/base.txt
robert-w-gries: Ok sounds good. Do you mind adding a comment regarding the min version we require and why?

robert-w-gries (approved):

PR #178, opened 2020-11-24
Add token authentication for petition endpoint so people can call it …

georgehelman:
…externally


robert-w-gries (changes requested):
Can you add documentation regarding how to use the token authentication? Also, is it possible to add tests for this functionality?

georgehelman:
Closing. No one wants this anymore.

Issue #181, opened 2020-12-15
Form 285 - Charges should be sorted  by file number instead of date of offense


georgehelman:
Clerks have to pull these files from physical storage, and that is in file number order, so it's easier for the clerks if the charges are sorted by file number instead of date of offense.

georgehelman:
https://github.com/deardurham/dear-petition/pull/185

PR #185, opened 2021-01-02
Sort by file number

georgehelman:
https://github.com/deardurham/dear-petition/issues/181

### DIFF
--- a/dear_petition/petition/export/forms.py
+++ b/dear_petition/petition/export/forms.py
@@ -23,7 +23,7 @@ class PetitionForm(metaclass=abc.ABCMeta):
     def get_ordered_offense_records(self):
         qs = self.petition.offense_records.select_related("offense__ciprs_record")
         return qs.order_by(
-            "offense__ciprs_record__offense_date",
+            "offense__ciprs_record__offense_date__year",
             "offense__ciprs_record__file_no",
             "pk",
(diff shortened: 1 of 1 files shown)


dear_petition/petition/export/forms.py
```diff
@@ -23,8 +23,8 @@ def disposition_code(self, offense):
     def get_ordered_offense_records(self):
         qs = self.petition.offense_records.select_related("offense__ciprs_record")
         return qs.order_by(
-            "offense__ciprs_record__offense_date",
             "offense__ciprs_record__file_no",
+            "offense__ciprs_record__offense_date",
```
copelco, line 27: Since ``file_no`` will sort alphanumeric, ``99CR------`` will sort after ``02CR------``, which will make the 1999 record show after the 2002 record. I think we could use ``CASE`` and ``WHEN`` to annotate the queryset with something like:
* ``year_sort=0`` WHERE offense_date < 2000
* ``year_sort=1`` WHERE offience_date >= 2000

If we sort by ``year_sort`` first, I think it'll work. 
* e.g. ``order_by("year_sort", "offense__ciprs_record__file_no")``

copelco (changes requested):

copelco (approved):

PR #189, opened 2021-01-16
Form 287 Box 3b

### LINKED ISSUE #180: Form 287 - Check mark in checkbox 3b
287, box 3B, “charged with multiple offenses, and while some charges were dismissed, others resulted in conviction” - when tool sees guilty to lesser, to pick up the charge from ciprs that says convicted

Eg. person charged with larceny, plea would be “guilty to lesser”, charge would be guilty and that should go into box 3B. New form also requires listing of remaining guilty charge.

copelco:
TODO
* [ ] Check the box

Closes #180


dear_petition/petition/export/forms.py
```diff
@@ -138,6 +140,22 @@ def map_offenses(self):
             self.data[f"DOOF:{i}"] = self.format_date(ciprs_record.offense_date)
             self.data[f"DismissalDate:{i}"] = self.format_date(offense.disposed_on)
 
+    def map_same_day_offenses(self):
+        dismissed_records = self.petition.get_all_offense_records()
+        charges = dismissed.same_day_convictions(dismissed_records)
+        line = []
+        for charge in charges[:1]:
```
robert-w-gries, line 147: Since the logic for both the first charge and the rest are almost identical maybe we can do this instead:

```python
        for index, charge in enumerate(charges):
            formatted = f"{charge.offense.ciprs_record.file_no} {charge.description.title()}"
            if index == 0:
                self.data["ChargedDesc"] = formatted
            else:
                self.data["ChargedDescCont"].append(f" {formatted}")
```

dear_petition/petition/export/tests/test_cr287_3b.py
```diff
@@ -0,0 +1,87 @@
...
+        county=county,
+    )
+    guilty_offense = OffenseRecordFactory(
+        action="CONVICTED",
+        description="BREAK OR ENTER A MOTOR VEHICLE",
+        severity="FELONY",
+        offense=OffenseFactory(
+            verdict="GUILTY",
+            disposition_method="DISPOSED BY JUDGE",
+            disposed_on=timezone.now(),
```
robert-w-gries, line 32: This is a fringe edge case, but with multiple calls to `timezone.now()` there's a very narrow window for the offense to be on separate days, like if we run the testcase milliseconds before midnight. I'll leave it up to you, but we could make a `time_disposed` variable that stores `timezone.now()` and pass it as the `disposed_on` value for each offense and offense record. Not a high priority change, so up to your discretion.

dear_petition/petition/export/tests/test_cr287_3b.py
```diff
@@ -0,0 +1,87 @@
...
+def test_same_day_convictions(durham_offense, durham_petition):
+    offenses = dismissed.same_day_convictions(durham_petition.get_all_offense_records())
+    assert durham_offense in offenses
+
+
+def test_cr287_same_day_convictions(durham_offense, durham_petition):
+    form = AOCFormCR287(durham_petition)
+    form.map_same_day_offenses()
+    expected = f"{durham_offense.offense.ciprs_record.file_no} {durham_offense.description.title()}"
+    assert form.data["ChargedDesc"] == expected
```
robert-w-gries, line 87: Can we add a testcase for `ChargedDescCont`?

robert-w-gries (changes requested):

dear_petition/petition/types/dismissed.py
```diff
@@ -28,3 +28,35 @@ def build_guilty_to_lesser_query():
...
+        OffenseRecord.objects.filter(pk__in=list(dismissed_record_ids))
+        .values_list("offense__ciprs_record", "offense__disposed_on")
+        .distinct()
+    )
+    records = []
+    for ciprs_record_id, disposed_on in qs:
+        guilty_records = OffenseRecord.objects.filter(
+            offense__verdict="GUILTY",
+            offense__ciprs_record__id=ciprs_record_id,
+            offense__disposed_on=disposed_on,
```
georgehelman, line 55: According to Jessica, we only want to check checkbox 3b if the guilty charge is on the same file number as a dismissed charge, so I don't think this logic is quite right.

dear_petition/petition/export/forms.py
```diff
@@ -138,6 +140,22 @@ def map_offenses(self):
...
+            line.append(
+                f"{charge.offense.ciprs_record.file_no} {charge.description.title()}"
+            )
+        self.data["ChargedDesc"] = " ".join(line)
+        line = []
+        for charge in charges[1:]:
+            line.append(
+                f"{charge.offense.ciprs_record.file_no} {charge.description.title()}"
+            )
+        self.data["ChargedDescCont"] = " ".join(line)
```
georgehelman, line 157: I am a bit confused by what all of this is doing. What is ChargedDesc and ChargedDescCont? I don't see those fields anywhere. I'm also not seeing the Checkbox annotation being used anywhere. Did the description of the issue change?

dear_petition/petition/types/dismissed.py
```diff
@@ -28,3 +28,35 @@ def build_guilty_to_lesser_query():
...
+            action="CONVICTED",
+        )
+        if guilty_records.exists():
+            records.extend(
+                guilty_records.select_related("offense__ciprs_record__batch")
```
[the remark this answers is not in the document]
copelco: [select_related](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#select-related) follows related relationships with SQL `JOIN`s, reducing future individual SQL queries while access related objects. For example, if the consumer of this function iterated over charges with `charge.offense.ciprs_record.file_no`, an individual SQL call would occur for each relation, offense and ciprs_record. With `select_related`, we capture everything in a single SQL query and improve performance.

dear_petition/petition/types/dismissed.py
copelco: I think that's what `offense__ciprs_record__id=ciprs_record_id` is doing, right?

dear_petition/petition/types/dismissed.py
```diff
@@ -28,3 +28,35 @@ def build_guilty_to_lesser_query():
...
+def same_day_convictions(dismissed_records):
+    """
+    To support field 3(b) on the AOC-CR-287 form, return a list of guilty
+    OffenseRecords that also had dismissals on the same day.
+
+    Petition language:
+        I was charged with multiple offenses, and while the charges
+        listed above were dismissed, the following charges resulted in a
+        conviction on the day of the dismissal or had not yet reached final
+        disposition.
```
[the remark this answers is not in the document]
copelco: Good question, I'm not sure Jessica has mentioned this yet.

dear_petition/petition/types/dismissed.py
```diff
@@ -28,3 +28,35 @@ def build_guilty_to_lesser_query():
...
+    """
+    dismissed_record_ids = dismissed_records.values_list("id", flat=True)
+    qs = (
+        OffenseRecord.objects.filter(pk__in=list(dismissed_record_ids))
```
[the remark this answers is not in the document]
copelco: I actually had that exact query before 9043bcaf7a6b474b65f55477f7104be7bc53f065. IIRC, since `dismissed_records` is sourced from `get_all_offense_records()`, which itself is a complex query, `distinct()` wasn't working properly. It's been a few months since I wrote it though... so I don't fully recall.

PR #198, opened 2021-03-13
Use GitHub actions

copelco:
Replace TravisCI with GitHub actions


.github/workflows/test.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Cache pip
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements/**.txt') }}
```
robert-w-gries, line 36: Seeing all the requirements files getting hashed made me curious.. Will the `pytest` environment test with `local` requirements or `production` requirements? If it's only testing using only one of these requrirements files, is this considered a gap in our testing infra?

robert-w-gries (approved):
Overall, looks good to merge. Had a question regarding our testing infra that likely doesn't apply to this PR

.github/workflows/test.yml
copelco: Currently, only `local.txt` is installed below, but the cache is invalidated even if a production requirement changes. I suppose we could install the production requirements too below. Seems like a good idea to me.

Issue #184, opened 2021-01-02
Check checkbox for AOC-285 when there is an AOC-285 attached


georgehelman:
The checkbox for the additional offenses form on the 287 and 288 form should be checked when there should be a corresponding 285 form attached.

![picture](https://user-images.githubusercontent.com/43391981/103450883-8fff4080-4c8a-11eb-9855-134fd14fb1f4.png)

georgehelman:
https://github.com/deardurham/dear-petition/pull/199

PR #199, opened 2021-03-15
Check additional petitions checkbox when there are additional 285 forms

georgehelman:
https://github.com/deardurham/dear-petition/issues/184

### DIFF
--- a/bin/get_annotations.py
+++ b/bin/get_annotations.py
@@ -1,0 +1,27 @@
+import sys
+
+import click
+import pdfrw
+
+
+@click.command()
+@click.argument("template")
+def get_annotations(template):
+    """
+    This utility prints all the annotations in the provided template.
+        template: A path to the template that you would like to get annotations for.
+    """
+
+    pdf = pdfrw.PdfReader(template)
+    annotations = pdf.pages[0]["/Annots"]
+    print("Here is a list of annotations in the document.\n")
+    try:
+        while True:
+            annotation = annotations.pop()
+            annotation_name = annotation['/T']
+            print(annotation_name)
+    except IndexError:
+        return
+
+if __name__ == '__main__':
+    sys.exit(get_annotations())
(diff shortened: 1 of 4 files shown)


bin/get_annotations.py
```diff
@@ -0,0 +1,27 @@
...
+    This utility prints all the annotations in the provided template.
+        template: A path to the template that you would like to get annotations for.
+    """
+
+    pdf = pdfrw.PdfReader(template)
+    annotations = pdf.pages[0]["/Annots"]
+    print("Here is a list of annotations in the document.\n")
+    try:
+        while True:
+            annotation = annotations.pop()
```
robert-w-gries, line 20: Why are you using a `while True` loop and catching an `IndexError` instead of iterating over the keys using `for annotation in annotations.values()`?

dear_petition/petition/export/forms.py
```diff
@@ -138,6 +139,10 @@ def map_offenses(self):
             self.data[f"DOOF:{i}"] = self.format_date(ciprs_record.offense_date)
             self.data[f"DismissalDate:{i}"] = self.format_date(offense.disposed_on)
 
+    def map_additional_forms(self):
+        if self.petition.has_attachments():
+            self.data[f"CkBox_Attchmt"] = Checkbox("Yes")
```
robert-w-gries, line 144: Is the format string necessary for `CkBox_Attchmt`?

bin/get_annotations.py
georgehelman: It doesn't have keys or values or anything useful like that unfortunately. It just has pop.

dear_petition/petition/export/forms.py
georgehelman: No, I'll remove that.

bin/get_annotations.py
robert-w-gries: `pop()` in this case is an array method. I ran the command locally and confirmed that `annotations` is an array. Maybe there was a typo when you tried last time?

robert-w-gries:
I thought of another addition you could make. Can you add a section to the readme listing how to use this tool? I personally did it by doing `docker-compose run django python get_annotations.py dear_petition/static/templates/AOC-CR-285.pdf`, not sure if that's how you intended

robert-w-gries (approved):

PR #218, opened 2021-04-14
Add celery to project

georgehelman:


config/settings/base.py
```diff
@@ -260,6 +265,14 @@
 # http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-soft-time-limit
 # TODO: set to whatever value is adequate in your circumstances
 CELERYD_TASK_SOFT_TIME_LIMIT = 60
+
+CELERY_BEAT_SCHEDULE = {
+    "clean_stale_data": {
+        "task": "dear_petition.petition.tasks.clean_stale_data.clean_stale_data_task",
+        "schedule": crontab(minute=0, hour="2"),
```
robert-w-gries, line 272: Does this run every day at 2 AM? Can you add a comment to indicate when it's expected to run?

dear_petition/celery.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from celery import Celery
+
+# set the default Django settings module for the 'celery' program.
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.local")
```
robert-w-gries, line 6: Are there different settings for non-local environments we should be using?

robert-w-gries (commented):
Couple minor questions but overall looks good to me. Colin might want to take a look as he likely has more Celery experience than me.

Liked the test addition btw!

config/settings/base.py
georgehelman: Sure. Yes, this schedules it to run at 2 AM EST each night.

dear_petition/celery.py
georgehelman: Yeah, it will be overwritten by Heroku config vars.

copelco (approved):

PR #237, opened 2021-07-05
Add underaged convictions petition

georgehelman:


dear_petition/petition/export/forms.py
```diff
@@ -28,7 +31,7 @@ def get_ordered_offense_records(self):
         # When sorting these, need to interpret first 2 digits of file number as year and sort based on that
         two_digit_current_year = timezone.now().year % 2000  # Returns 21 given 2021
         qs = (
-            self.petition.offense_records.select_related("offense__ciprs_record")
+            self.petition.offense_records.filter(active=True)
```
robert-w-gries, line 34: Why did you remove the `select_related()` call?

dear_petition/petition/types/tests/test_underaged_convictions.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_underaged_conviction(batch, record1, non_dismissed_offense):
+    today = timezone.now()
+    record1.dob = record1.offense_date.date() - timedelta(days=1)
+    record1.save()
+    offense_record = OffenseRecordFactory(
+        action="CONVICTED", offense=non_dismissed_offense
+    )
+    assert offense_record in batch.underaged_conviction_records()
```
robert-w-gries, line 24: Can you add a test to ensure that someone who is one day away from their 18th birthday will also be marked as underaged?

src/components/elements/HighlightTable/HighlightTable.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import styled from 'styled-components';
+import { TableBody, TableCell, TableHeader, TableRow } from '../Table';
+import { colorPrimary } from '../../../styles/colors';
+
+const TableStyle = styled.table`
```
robert-w-gries, line 6: We're repeating a lot css here. We create a base `TableStyled` styled-component and extend it as necessary by doing `const HighlightTableStyle = styled(TableStyle)`

src/components/elements/HighlightTable/HighlightTable.js
```diff
@@ -0,0 +1,109 @@
...
+            <TableCell>{offenseRecord.action}</TableCell>
+            <TableCell>{offenseRecord.severity}</TableCell>
+            <TableCell>{offenseRecord.law}</TableCell>
+        </TableRow>
+    )
+};
+
+function HighlightTable({ offenseRecords, highlightRow, highlightedRows, unhighlightRow, setIsModified }) {
+    return (
+        <StyledTable numColumns={8} columnSizes='.1fr .4fr .4fr 1.2fr .8fr .4fr .4fr .6fr'>
```
robert-w-gries, line 82: I don't think we should use `fr` like this. A unit of `fr` already represents one "part" of the width/height of the container. Adding fractions to what is already a fraction makes it more difficult to understand what the relative width of each column should be.

We should only use whole numbers. No matter how large the `fr` size is, the total width will always sum up to 100%. So you can do something like this:

`columnSizes='1fr 4fr 4fr 12fr 8fr 4fr 4fr 6fr'`

src/components/elements/Table.js
```diff
@@ -40,24 +40,43 @@ const TableStyle = styled.table`
...
 `;
 
-export const TableCell = ({ children, header }) => (
+export const TableCell = ({ children, header, }) => (
   <>{header ? <th>{children}</th> : <td>{children}</td>}</>
 );
 
-export const TableBody = styled.tbody``;
+export const TableSpanCell = styled.td`
+  grid-column: 1 / span 5;
```
robert-w-gries, line 50: Can you make the `span` a prop so we can re-use this styled component?

src/components/elements/Table.js
```diff
@@ -40,24 +40,43 @@ const TableStyle = styled.table`
...
+`
+
+export const TableRightAlignCell = styled.td`
+  text-align: right;
+  margin-right: 50px;
```
robert-w-gries, line 55: Using text-align + margin-right shouldn't be necessary since we're using css grid. It doesn't look like this styled component is being used anyway, so can you remove it?

src/components/elements/Table.js
```diff
@@ -40,24 +40,43 @@ const TableStyle = styled.table`
...
+`
+
+export const TableBody = ({children}) => (
+  <tbody>{children}</tbody>
```
robert-w-gries, line 59: Not sure there's any benefit to this diff. Were you planning to extend the `TableBody` component? I'd rather keep the html `<table>` specific stuff as passthrough elements:
```
export const TableBody = styled.tbody``;
```

src/components/elements/Table.js
```diff
@@ -40,24 +40,43 @@ const TableStyle = styled.table`
...
+  &>td {
+  ${props => props.highlighted ? `
+  background-color: rgb(255, 245, 217);
+  ` : `background-color: ${props.backgroundColor}`}
+  }
+  cursor: pointer;
+
+  input[type="checkbox"] {
+    cursor: pointer;
+  }
```
robert-w-gries, line 78: Is this style block doing anything?

src/components/pages/DetailRecordPage/DetailRecordPage.js
```diff
@@ -0,0 +1,159 @@
...
+function PageButton({ label, onClick, disabled }) {
+    return (
+      <ButtonStyled onClick={onClick} disabled={disabled}>
+        {label}
+      </ButtonStyled>
+    );
+  }
+
+
+function DetailRecordPage() {
```
robert-w-gries, line 38: Just want to make sure. You removed the route to this page, right? I think we'll eventually use this separate page, but that should be in a subsequent update

src/components/pages/GenerationPage/GenerationPage.js
```diff
@@ -5,7 +5,7 @@ import { colorGrey } from '../../../styles/colors';
 import { smallerThanTabletLandscape } from '../../../styles/media';
 
 // Router
-import { useParams } from 'react-router-dom';
+import { useParams, } from 'react-router-dom';
```
robert-w-gries, line 8: Nitpick:  unnecessary comma here

src/components/pages/GenerationPage/PetitionList.js
```diff
@@ -51,9 +67,43 @@ function PetitionRow({ attorney, petition, petitionerData, validateInput }) {
...
+  }
+
+  const unhighlightRow = ( offenseRecordId ) => {
+    setHighlightedRows(highlightedRows.filter((value, index, arr) => {return value !== offenseRecordId}));
+  }
+
+  const recalculatePetitions = () => {
+    Axios.post(`/petitions/${petitionId}/recalculate_petitions/`, {"offense_record_ids": highlightedRows}).then(
+      ({ data }) => {
+        console.log(data);
```
robert-w-gries, line 96: Nitpick: Can you remove this `console.log()`?

src/components/pages/GenerationPage/PetitionList.js
```diff
@@ -79,7 +131,43 @@ function PetitionRow({ attorney, petition, petitionerData, validateInput }) {
...
       </TableRow>
+      {isDetailed && (
+        <TableRow key={petition.pk + 'offenses'} backgroundColor={backgroundColor}>
+          <TableSpanCell className='spanCell'>
+            { offenseRecordsLoading ? <h5>Loading...</h5> : (
+              <Table >
+                <TableBody>
+                  <TableRow key={petition.pk + '_offenses'}>
+                    <TableCell>
+                        <HighlightTable
```
robert-w-gries, line 151: This doesn't look right to me. Why are there three nested tables here? I think we can simplify:

```
<TableSpanCell>
  <FlexColumnDiv>
    <HighlightTable />
    <AlignSelfEndGenerateButton />
  </FlexColumnDiv>
</TableSpanCell>
```

Note: you'll need to add `align-self: end` or something like that to the generate button and you'll need to add a wrapper element with `display: flex` and `flex-flow: column`

src/components/pages/GenerationPage/PetitionList.js
```diff
@@ -79,7 +131,43 @@ function PetitionRow({ attorney, petition, petitionerData, validateInput }) {
...
+            label={isDetailed ? <FontAwesomeIcon icon={faChevronDown}/> : <FontAwesomeIcon icon={faChevronRight}/>}
+            isCollapsed=<FontAwesomeIcon icon={faChevronDown}/>
+            onClick={() => handlePress()}
+            title="Reveal offense records"
+          />
+        </TableCell>
       </TableRow>
+      {isDetailed && (
+        <TableRow key={petition.pk + 'offenses'} backgroundColor={backgroundColor}>
+          <TableSpanCell className='spanCell'>
```
robert-w-gries, line 145: Remove the className

src/components/pages/GenerationPage/PetitionList.js
```diff
@@ -79,7 +131,43 @@ function PetitionRow({ attorney, petition, petitionerData, validateInput }) {
...
+          <GenerateButton 
+            label={isDetailed ? <FontAwesomeIcon icon={faChevronDown}/> : <FontAwesomeIcon icon={faChevronRight}/>}
+            isCollapsed=<FontAwesomeIcon icon={faChevronDown}/>
+            onClick={() => handlePress()}
+            title="Reveal offense records"
+          />
+        </TableCell>
       </TableRow>
+      {isDetailed && (
+        <TableRow key={petition.pk + 'offenses'} backgroundColor={backgroundColor}>
```
robert-w-gries, line 144: You don't need `key` prop unless you're using something like `list.map(() => <div key={key} />`

https://reactjs.org/docs/lists-and-keys.html

robert-w-gries (changes requested):
Nice work! Before approving, please address the comments and also fix the prettier/eslint messages. They're currently blocking me from running your branch locally (are you using localhost:3000 or localhost:8000? port 3000 is the correct one for local dev)

dear_petition/petition/api/serializers.py
```diff
@@ -30,9 +30,34 @@ def get_admin_url(self, obj):
 
 
 class OffenseRecordSerializer(serializers.ModelSerializer):
+    offense_date = serializers.SerializerMethodField()
+    dob = serializers.SerializerMethodField()
+    disposition_method = serializers.SerializerMethodField()
+
+    def get_offense_date(self, offense_record):
+        return offense_record.offense.ciprs_record.offense_date.date()
```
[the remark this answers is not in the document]
georgehelman: I don't think it's significant. The other options would be to denormalize the offense_date field or cache it per API query somehow.

dear_petition/petition/api/serializers.py
```diff
@@ -76,6 +101,10 @@ class Meta:
 
 class PetitionSerializer(serializers.ModelSerializer):
     jurisdiction = serializers.CharField(source="get_jurisdiction_display")
+    offense_records = serializers.SerializerMethodField()
+
+    def get_offense_records(self, petition):
+        return petition.offense_records.filter(active=True).values_list("id", flat=True)
```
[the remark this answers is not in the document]
georgehelman: Yes, it filters active records. I don't think this endpoint was used by the frontend previously but let me know if that's wrong. I'll use get_all_offense_records

src/components/elements/Table.js
georgehelman: Yeah, it makes the cursor a pointer when hovering the checkbox. Idk why it's needed but it is.

dear_petition/petition/api/viewsets.py
```diff
@@ -44,6 +46,33 @@ class OffenseRecordViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.OffenseRecordSerializer
     permission_classes = [permissions.IsAuthenticated]
 
+    @action(
+        detail=False,
+        methods=[
+            "get",
+        ],
+    )
+    def get_petition_records(self, request):
```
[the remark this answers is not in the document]
georgehelman: Maybe it should be. Right now, I request this data when you open the offense record table panel, but maybe I should just include this data in the original batch API call when the page is loaded. I'll try it.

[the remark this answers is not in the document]
georgehelman: I think it's fine the way it is. It's returning offense records. It should be in the offense record viewset.

src/components/pages/DetailRecordPage/DetailRecordPage.js
georgehelman: Yes, I removed the route.

dear_petition/petition/api/serializers.py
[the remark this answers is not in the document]
robert-w-gries: It impacts the Batch serializers, which I think may want to know all the offense records, even if they're not "active". I think it'd be better to have the `offense_records` field return all offense records and use a different, specialized function to filter for active when we need it

dear_petition/petition/api/viewsets.py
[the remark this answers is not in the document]
robert-w-gries: It just seems like unnecessary overhead to me to make a separate route in OffenseRecordViewset with `detail=False` to act like we're working on a list of every OffenseRecord before we just use the petition id from the request to retrieve a different View and work with the offense records for that specific petition.

In my mind you don't even need to create a special route/action. You can just make the `/petitions/${petitionId}` return the petition object with the fields you specified above and even more. I could see that route being very useful in general.

This isn't a blocker for merging, so we can spin this out to a different issue. If you want to discuss it more, we can debate it in that ticket

dear_petition/petition/api/serializers.py
```diff
@@ -90,12 +119,22 @@ class Meta:
...
 
     def get_attachments(self, instance):
-        return PetitionSerializer(instance.attachments.all().order_by('pk'), many=True).data
+        return PetitionSerializer(
+            instance.attachments.all().order_by("pk"), many=True
+        ).data
+
+    def get_offense_records(self, petition):
+        return list(
+            petition.get_all_offense_records(filter_active=False).values_list(
```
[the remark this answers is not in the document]
robert-w-gries: You resolved this comment but I don't see a commit addressing it? Are you planning to move forward without changing the `get_attachments` function?

src/components/elements/Table.js
robert-w-gries: You marked as resolved but didn't change it. Was this intentional?

src/components/elements/Table.js
robert-w-gries: You marked as resolved but didn't change it. Was this intentional?

Issue #265, opened 2022-01-19
Start tracking zip code, race, gender, age in Generated Petition


georgehelman:

georgehelman:
https://github.com/deardurham/dear-petition/pull/273

PR #273, opened 2022-01-23
Add age, county, jurisdiction, race, sex, and county to tracked metrics

georgehelman:


georgehelman:
https://github.com/deardurham/dear-petition/issues/265

dear_petition/petition/models.py
```diff
@@ -332,3 +346,12 @@ class GeneratedPetition(TimeStampedModel):
     form_type = models.CharField(choices=FORM_TYPES, max_length=255)
     number_of_charges = models.IntegerField()
     batch_id = models.PositiveIntegerField()
```
copelco, line 348: I'd suggest making all of the new fields read-only in the admin, so these values can't be modified.

robert-w-gries (changes requested):
Is there any testing for the statistics yet? I think that could be a good addition if it won't take too much time. Otherwise, we can split it out to a new ticket

robert-w-gries (approved):

georgehelman:
I'll split it into a new issue.

georgehelman:
https://github.com/deardurham/dear-petition/issues/280

Issue #264, opened 2022-01-19
Allow generated petitions to be exported to CSV by admins


georgehelman:
This can be a button on the admin panel.

PR #274, opened 2022-01-30
#264 Allow generated petitions to be downloaded as CSV

georgehelman:


dear_petition/petition/api/viewsets.py
```diff
@@ -245,6 +247,62 @@ def create(self, request):
         return resp
 
 
+def download_csv(modeladmin, request, queryset):
```
copelco, line 250: Is this non-class `download_csv` method used anywhere?

copelco (commented):
Added a comment. Otherwise works for me.

dear_petition/petition/api/viewsets.py
georgehelman: Ah, no. I stopped and started a little bit while working on this. Good catch.

copelco (approved):

PR #284, opened 2022-02-23
Create separate sentry envs for different envs

georgehelman:


copelco (commented):
Is it possible to set the necessary environment variables to handle the config differences? I believe a single settings file configured via env vars will be easier to manage in the long run.

PR #290, opened 2022-03-15
Create m2m model for petitions and offense records to correctly track active records

georgehelman:
The new PetitionDocument model makes it possible to have an offense record be active on one document and inactive on another document of a different petition type by separating the concept of the overarching petition object and the documents (including attachments) associated with that object. Previously all this data was stored in a single table, which is terribly awkward because they are really two different concepts. This not only solves the current bug but is an improvement to the data model overall.


dear_petition/petition/models.py
```diff
@@ -339,13 +337,51 @@ def get_all_offense_records(self, filter_active=True, include_annotations=True):
...
     def has_attachments(self):
-        return self.attachments.count() > 0
+        return self.documents.count() > 1
 
 
 # Look-alike Petition object used to support JSON data-driven petitions
 DataPetition = namedtuple("DataPetition", ["form_type", "data_only"], defaults=[True])
 
 
+class PetitionDocument(models.Model):
```
copelco, line 347: What does `PetitionDocument` provide over `Petition`? `Petition` provides attachments and linked offense records (now through `PetitionOffenseRecord`).

dear_petition/petition/models.py
```diff
@@ -339,13 +337,51 @@ def get_all_offense_records(self, filter_active=True, include_annotations=True):
...
+    @property
+    def jurisdiction(self):
+        return self.petition.jurisdiction
+
+    @property
+    def county(self):
+        return self.petition.county
+
+
+class PetitionOffenseRecord(models.Model):
```
copelco, line 379: Should this be used by `PetitionDocument.offense_records`?

copelco (commented):
I took a quick pass through tonight. The CI build is currently failing.

I'm not sure I understand the motivation for the addition of `PetitionDocument`. Do you mind updating the PR description with a summary of what this accomplishes? Sorry if I'm missing something obvious!

dear_petition/petition/models.py
georgehelman: I updated the description to explain.

dear_petition/petition/models.py
```diff
@@ -339,13 +337,51 @@ def get_all_offense_records(self, filter_active=True, include_annotations=True):
...
+    @property
+    def is_attachment(self):
+        return self.previous_document is not None
+
+    @property
+    def form_type(self):
+        if self.previous_document_id:
+            return pc.ATTACHMENT
+        else:
+            return self.petition.form_type
```
copelco, line 368: Since we have this, can we delete the Petition.parent field?

dear_petition/petition/models.py
copelco, line 379: It might be nice to have an ModelAdmin for this model

dear_petition/__init__.py
```diff
@@ -1,11 +0,0 @@
...
-
-__version__ = "0.1.0"
-__version_info__ = tuple(
-    [
-        int(num) if num.isdigit() else num
-        for num in __version__.replace("-", ".", 1).split(".")
-    ]
-)
-
-__all__ = ("celery_app",)
```
copelco, line 11: Hmm, do we not need `celery_app` anymore? I think that's how Celery finds the periodic tasks.

copelco (approved):
Works for me!

Issue #293, opened 2022-03-29
Create common dockerfile for prod and dev builds


robert-w-gries:
Use multi-stage builds with "development" or "production" target to add on the extra software

PR #303, opened 2022-05-27
Use multi-stage build for Django dev environment

copelco:
### Issue
- #293

### Testing
Run `docker-compose up --build -d django`
Visit http://localhost:8000/


requirements/base.txt
```diff
@@ -25,4 +25,4 @@ djangorestframework==3.13.1  # https://github.com/encode/django-rest-framework
 djangorestframework-simplejwt==5.2.0 # https://github.com/SimpleJWT/django-rest-framework-simplejwt
 
 pdfrw==0.4
--e git+https://github.com/deardurham/ciprs-reader.git@main#egg=ciprs-reader
+git+https://github.com/codefordurham/ciprs-reader.git@main#egg=ciprs-reader
```
robert-w-gries, line 28: Was this intentional? I don't think we should use the old location here

src/components/elements/Input/FormTextArea.jsx
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-// TODO: why doesn't the above rule handle wrapped textarea properly?
```
robert-w-gries, line 2: Not sure I understand why this was removed. You said you were seeing an eslint error and somehow removing the eslint-disable fixed it?

docker-compose.yml
```diff
@@ -53,12 +44,6 @@ services:
       - ./.envs/.local/.frontend
     environment:
       - OVERRIDE_API_PROXY
-      - NODE_PATH=/home/node_install/node_modules
-
-  mailhog:
-    image: mailhog/mailhog:v1.0.0
-    ports:
-      - "8025:8025"
```
robert-w-gries, line 61: Why was mailhog removed? I thought we needed this to test mail locally

docker-compose.deploy.yml
```diff
@@ -1,15 +1,15 @@
...
+    build:
+      context: .
+      target: deploy
     env_file:
-      - .envs/.production/.django
-    image: petition:latest
-    links:
-      - db:db
-      - redis:redis
+      - .envs/.production-local-testing/.django
```
robert-w-gries, line 9: This env file name seems off to me. Does the actual deploy ignore/omit this env file? Is there a way to make sure we don't include such a file in the deploy?

README.md
```diff
@@ -69,6 +92,27 @@ Run the containers:
...
+  django:
+    command: ["sleep", "infinity"]
+```
+
+Now we run `runserver` manually to have more control over restarts:
+
+```sh
+docker compose exec django bash
+root$ python manage.py runserver 0.0.0.0:8000
+```
```
robert-w-gries, line 114: I don't think i'm fully understanding this section. Is this a new required step in getting the project to run? Is it just a suggestion?

robert-w-gries (changes requested):

requirements/base.txt
copelco: Wow, good catch, how did that happen 😮 -- fixed

src/components/elements/Input/FormTextArea.jsx
copelco: The node build fails with the following error, I probably just don't know the right way to fix it:

```
❯ docker-compose up --build -d django
...
 => CACHED [static_files 2/8] WORKDIR /code                                                                                                                                                       0.0s
 => CACHED [static_files 3/8] COPY package.json package-lock.json tailwind.config.js /code/                                                                                                       0.0s
 => CACHED [static_files 4/8] RUN npm install --silent                                                                                                                                            0.0s
 => CACHED [static_files 5/8] COPY ./public /code/public/
 => CACHED [static_files 6/8] COPY ./src /code/src/                                                                                                                                               0.0s
 => CACHED [static_files 7/8] WORKDIR /code/
 => ERROR [static_files 8/8] RUN npm run build                                                                                                                                                   17.8s
------                                                                                                                                                                                                 
 > [static_files 8/8] RUN npm run build:                                                                                                                                                               
#26 0.518                                                                                                                                                                                              
#26 0.518 > dearpetition@0.1.0 build                                                                                                                                                                   
#26 0.518 > react-scripts build                                                                                                                                                                        
#26 0.518 
#26 1.817 Creating an optimized production build...
#26 17.69 Failed to compile.
#26 17.69 
#26 17.69 src/components/elements/Input/FormTextArea.jsx
#26 17.69   Line 1:1:  Definition for rule 'jsx-a11y/label-has-associated-control' was not found  jsx-a11y/label-has-associated-control
#26 17.69 
#26 17.69 Search for the keywords to learn more about each error.
#26 17.69 
#26 17.69 
------
executor failed running [/bin/sh -c npm run build]: exit code: 1
ERROR: Service 'django' failed to build : Build failed
```

docker-compose.yml
copelco: Local development now uses the console EmailBackend, so messages are just printed to stdout in the Django container:

https://github.com/deardurham/dear-petition/blob/5f2bce75d5efc22761ffa07de0aaa076777aa3dd/config/settings/local.py#L35

docker-compose.deploy.yml
copelco: Heroku and the production deploy don't use this file. It's only used to test building and running the `deploy` stage of the Dockerfile.

And yes, this file isn't committed and is in gitignore:

https://github.com/deardurham/dear-petition/blob/5f2bce75d5efc22761ffa07de0aaa076777aa3dd/.gitignore#L287-L288

README.md
copelco: It's a suggestion. By default, `runserver` starts automatically in the `dev` stage:

https://github.com/deardurham/dear-petition/blob/8eb977bc56fdaa8bb4b88f914d161e90a8a4282c/Dockerfile#L150

However, this doesn't always fail cleanly when say a syntax error is introduced in a Python module and the file watcher attempts to auto-reload runserver. It may fail entirely and you have to re-run `docker compose up -d django up`. If you use the suggestion documented here though, the user manually runs runserver within the container, so if it fails, you can just restart the process.

georgehelman (approved):

PR #309, opened 2022-07-04
Bring agencies into petition document generation process

georgehelman:


dear_petition/petition/etl/load.py
```diff
@@ -93,3 +92,42 @@ def create_documents(petition):
...
+
+    try:
+        current_document = current_document.following_document
+        if not current_document.offense_records.exists():
+            # This must have been an extra attachment for agencies, but agencies were removed so no longer necessary
+            current_document.delete()
+    except:
+        pass
+
+    return petition
```
georgehelman, line 133: This does not handle edge case where an existing attachment needs to have it's agency cleared

robert-w-gries (approved):

Issue #326, opened 2022-10-20
Create "my inbox" REST endpoint


copelco:

copelco:
Implemented in #327

PR #327, opened 2022-10-21
Create My Inbox REST endpoint

copelco:
### Issue
- #326 

### Summary
- New `/petition/api/my-inbox/` API view returns list of a user's batches and related metadata:
  ```json
  {
      "count": 2,
      "next": null,
      "previous": null,
      "results": [
          {
              "pk": 19,
              "label": "janedoe",
              "date_uploaded": "2022-07-07T10:35:50.701216-04:00",
              "automatic_delete_date": "2022-10-23T21:42:26.330318-04:00",
              "total_files": 1,
              "total_emails": 0,
              "total_petitions": 4,
              "total_ciprs_records": 34
          },
          {
              "pk": 18,
              "label": "jondoe",
              "date_uploaded": "2022-07-07T10:35:50.701216-04:00",
              "automatic_delete_date": "2022-10-23T21:42:26.330318-04:00",
              "total_files": 3,
              "total_emails": 3,
              "total_petitions": 2,
              "total_ciprs_records": 18
          }
      ]
  }
  ```
- Create `Batch`<->`Email` and `CIPRSRecord`->`BatchFile` relationships, so PDFs can be stitched back together in the future
- Implement more generic `settings.SENDGRID_WEBHOOK_LISTENERS` workflow
- Secure webhook endpoint using optional `settings.SENDGRID_ALLOWED_SENDERS` setting
- Organize media files by environment (e.g. `production/batch/53/hello.pdf`, `staging/batch/21/foo.pdf`) so files don't collide in S3 bucket


dear_petition/petition/api/tests/test_viewsets.py
```diff
@@ -180,3 +180,25 @@ def test_export_as_csv(self):
...
         self.assertTrue(lines[1].startswith("1"))
+
+
+class TestMyInbox:
+    def test_login_required(self, api_client_anon):
+        response = api_client_anon.get(reverse("api:my-inbox"))
+        assert response.status_code == 401
+
+    @pytest.mark.django_db
+    def test_foo(self, user, api_client):
```
robert-w-gries, line 191: Was this test's name a placeholder?

dear_petition/petition/etl/email.py
```diff
@@ -0,0 +1,47 @@
...
+def sendgrid_webhook_listener(email_id):
+    """
+    Associate email attachments to a user and pass off to the ETL process.
+    """
+    email = Email.objects.get(id=email_id)
+    username, label = extract_username_and_label(email.recipient)
+    try:
+        user = User.objects.get(username=username)
+    except User.DoesNotExist:
+        logger.error(f"Username {username} not found for email '{email}'")
```
robert-w-gries, line 38: Just thinking.. is there anything else we should do in the case where we can't extract a username? Should we have a bin of unmatched emails that we clean out periodically? Would we be able to match emails that should have been mapped to a user/batch but weren't?

dear_petition/petition/etl/load.py
```diff
@@ -31,14 +31,11 @@ def import_ciprs_records(files, user, parser_mode, batch_label=""):
     logger.info(f"Created batch {batch.id}")
     for idx, file_ in enumerate(files):
         logger.info(f"Importing file {file_}")
-        if settings.CIPRS_SAVE_PDF:
```
robert-w-gries, line 34: Should we remove this setting if we're not using it here?

dear_petition/petition/etl/email.py
```diff
@@ -0,0 +1,47 @@
...
+        return
+    attachment_files = []
+    for attachment in email.attachments.all():
+        attachment_files.append(attachment.file)
+    batch = import_ciprs_records(
+        files=attachment_files, user=user, parser_mode=1, batch_label=label
+    )
```
robert-w-gries, line 45: The cool thing about using the email listener is that we can addsome some background processing on the ciprs record here without worrying about taking too much time!

If you agree it's a good idea, I think it'd be nice to open an issue detailing potential improvements. One I've thought about it adding image PDF parsing. Another is using something called Camelot which is much slower than xpdfreader but it has better support for parsing of tables.

dear_petition/petition/models.py
```diff
@@ -220,11 +230,22 @@ def age(self):
...
+
+def get_batch_file_upload_path(instance, filename):
+    return "batch/%s/%s" % (instance.batch.id, filename)
+
 
 class BatchFile(models.Model):
     batch = models.ForeignKey(Batch, related_name="files", on_delete=models.CASCADE)
     date_uploaded = models.DateTimeField(auto_now_add=True)
-    file = models.FileField(upload_to="ciprs/")
+    file = models.FileField(upload_to=get_batch_file_upload_path)
```
robert-w-gries, line 248: How is `filename` getting passed here? I feel like this is a django thing i'm not understanding

dear_petition/sendgrid/views.py
```diff
@@ -1,25 +1,39 @@
...
         return HttpResponse(status=400)
-    email = form.save()
-    link_email_to_user(email_id=email.id)
+    sender = form.cleaned_data["sender"]
+    allowed_senders = getattr(settings, "SENDGRID_ALLOWED_SENDERS", [])
+    if email_is_allowed(addr=sender, allowed_addrs=allowed_senders):
+        email = form.save()
+        notify_sendgrid_listeners(email_id=email.id)
+    else:
+        logger.warning(f"Email address not allowed {sender}")
```
robert-w-gries, line 38: Maybe return a 403 here?

dear_petition/petition/etl/email.py
copelco: `Batch.user` is currently required, so it's not possible using the current model, but the `Email` and `Attachment` objects will still exist even if a matching user isn't found. Logger errors are picked up by Sentry, so we will be notified. I think it's a good idea though that we can ask them about.

dear_petition/petition/models.py
copelco: It's passed to the function:

https://github.com/deardurham/dear-petition/blob/029e1533f9c626ffc07237c275062c81381948d3/dear_petition/petition/models.py#L241-L242

dear_petition/sendgrid/views.py
copelco: Anything other than a 2xx will cause SendGrid to retry for 3 days for a message. I'm not sure it makes sense here.. if we don't allow the email, we just don't save it.

dear_petition/petition/etl/load.py
copelco: Done!

dear_petition/petition/etl/email.py
robert-w-gries: Did you ask about this in the meeting? Curious if it came up

robert-w-gries (approved):

PR #343, opened 2022-11-27
Update readme

rebecca-draben:
Make instructions more clear for someone new to the project and possibly new to Docker too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,11 @@ A [Durham Expunction and Restoration (DEAR)](https://www.deardurham.org)
 git clone git@github.com:deardurham/dear-petition.git
 cd dear-petition
 docker-compose up -d django
+docker-compose run --rm django python manage.py migrate
 docker-compose run --rm django python manage.py createsuperuser
 ```
 
+When asked for a username and password, enter values of your choosing.  Email address may be left empty.
 
 ## Frontend Development
 
@@ -77,6 +79,10 @@ You can set the proxy url by either setting `OVERRIDE_API_PROXY` or `API_PROXY`:
 API_PROXY=http://localhost:8888 npm start
 ```
 
+### Try it out
+
+Try out DEAR Petition Generator by logging in as the superuser you created and uploading test CIPRS records (in pdf format) which can be downloaded from https://github.com/deardurham/ciprs-reader/tree/main/tests/test_records
+
 ## Backend Development (with Docker)
 
 To run this on a Mac, use [Docker for
@@ -90,12 +96,12 @@ Run the containers:
 
     docker-compose up django
 
-Visit http://localhost:8000/petition/api/ in your browser.
+Visit http://localhost:8000/petition/api/ in your browser.  If you get authentication errors, you may login as the superuser you created at http://localhost:8000/ and try again.
 
 ### Using docker-compose.override.yml
 
-To develop in a Docker container, we'll use a `docker-compose.override.yml`
-override file to configure the Django container to sleep by default:
+To develop in a Docker container, we'll create a `docker-compose.override.yml`
+override file in the root of the dear-petition directory to configure the Django container to sleep by default:
 
 ```yaml
 # file: docker-compose.override.yml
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,27 +115,30 @@ services:
...
 ```sh
-docker compose exec django bash
+docker-compose up -d django
+docker-compose exec django bash
 root$ python manage.py runserver 0.0.0.0:8000
 ```
 
 ### Initial Setup
 
+Migrate DB:
```
rebecca-draben, line 125: Switched the order of migrate and createsuperuser because you'll get fewer warnings in red.

README.md
```diff
@@ -109,27 +115,30 @@ services:
...
+When asked for a username and password, enter values of your choosing.  Email address may be left empty.    
 
 See detailed [cookiecutter-django Docker
 documentation](http://cookiecutter-django.readthedocs.io/en/latest/deployment-with-docker.html).
 
 
 ### Restore database
 
-To restore a database dump, you can run:
+If you have a database dump you wish to restore, you can run:
```
rebecca-draben, line 141: Slightly different wording might prevent someone who is quickly running through the readme from dropping the db when they don't have a database dump to use for the restore.

README.md
```diff
@@ -36,9 +36,11 @@ A [Durham Expunction and Restoration (DEAR)](https://www.deardurham.org)
 git clone git@github.com:deardurham/dear-petition.git
 cd dear-petition
 docker-compose up -d django
+docker-compose run --rm django python manage.py migrate
 docker-compose run --rm django python manage.py createsuperuser
 ```
 
+When asked for a username and password, enter values of your choosing.  Email address may be left empty.
```
robert-w-gries, line 43: I think this can be removed from the `Quick Start` section, as we already have the same listed in the ` Initial Setup` section

README.md
```diff
@@ -77,6 +79,10 @@ You can set the proxy url by either setting `OVERRIDE_API_PROXY` or `API_PROXY`:
 API_PROXY=http://localhost:8888 npm start
 ```
 
+### Try it out
+
+Try out DEAR Petition Generator by logging in as the superuser you created and uploading test CIPRS records (in pdf format) which can be downloaded from https://github.com/deardurham/ciprs-reader/tree/main/tests/test_records
```
robert-w-gries, line 84: Thanks for adding this. I'm not entirely sure where the proper place is to add it. I'm thinking the `Quick Start` section because that's likely the section most new developers would be reading

README.md
```diff
@@ -90,12 +96,12 @@ Run the containers:
...
 
-Visit http://localhost:8000/petition/api/ in your browser.
+Visit http://localhost:8000/petition/api/ in your browser.  If you get authentication errors, you may login as the superuser you created at http://localhost:8000/ and try again.
 
 ### Using docker-compose.override.yml
 
-To develop in a Docker container, we'll use a `docker-compose.override.yml`
-override file to configure the Django container to sleep by default:
+To develop in a Docker container, we'll create a `docker-compose.override.yml`
+override file in the root of the dear-petition directory to configure the Django container to sleep by default:
```
robert-w-gries, line 104: Your changes look good. I think this section could also probably be elaborated upon because it is currently portrayed as if it's a required step. Some ideas to de-prioritize this section:

* Add `(Optional)` to the header
* Rename the header to something like `Configuring the containers using docker-compose.override.yml`
* Move this section below the `Initial Setup` and `Restore database` sections

robert-w-gries (changes requested):

README.md
rebecca-draben: I'll remove it.  I was conflicted both by repeating it and by not having it both places.

README.md
rebecca-draben: Will do

README.md
rebecca-draben: Will use all three suggestions

robert-w-gries (approved):

PR #351, opened 2023-01-07
Remove __str__ from PrintableModelMixin

rebecca-draben:
The Django Admin UI uses __str__ extensively, so it is better to revert to the less verbose __str__ implementations


copelco (approved):
Admin looked good to me - thanks @rebecca-draben !

Issue #345, opened 2022-12-06
Attorney state/zip not populated on UI 


rebecca-draben:
When I search for and select an attorney, the attorney's contact info is populated, but not their state or zip.  I noticed this in the staging environment.  

The state and zip are populated in the PDF, so this is just a cosmetic issue. 

<img width="600" alt="Screen Shot 2022-12-06 at 6 19 52 PM" src="https://user-images.githubusercontent.com/1581141/206047110-11484e20-09ca-4a68-9608-4b74559fdab2.png">

<img width="600" alt="Screen Shot 2022-12-06 at 6 21 49 PM" src="https://user-images.githubusercontent.com/1581141/206047120-dc703cb4-534d-4d24-a393-ee166cd20571.png">

austin-bagwell:
Rebecca caught an error caused by this commit:

<img width="448" alt="Screen Shot 2023-02-03 at 5 39 28 PM" src="https://user-images.githubusercontent.com/113562625/216724674-4ab9a44b-d2c5-4113-b0f3-3fffe43a4432.png">

Error message:
<img width="888" alt="Screen Shot 2023-02-03 at 5 41 39 PM" src="https://user-images.githubusercontent.com/113562625/216724837-c24b273c-cc96-410a-bb30-b426646e8a45.png">

I added this ternary back into AddressInput:
<img width="973" alt="Screen Shot 2023-02-03 at 5 28 35 PM" src="https://user-images.githubusercontent.com/113562625/216724912-95410640-e171-4fb9-a434-6040690c5e62.png">

Which seems to resolve the error above. It does throw this error, but doesn't break the page:
![Screen Shot 2023-02-03 at 5 30 11 PM](https://user-images.githubusercontent.com/113562625/216724985-7f489ab8-097b-4c49-999c-21844bcc2a98.png)

Would it be best to create a new PR with that ternary added back, or revert master back to the version prior to my PR?

robert-w-gries:
That warning in the fourth screenshot is unrelated. You can create a new PR to fix the error

austin-bagwell:
Just opened a new PR. sorry about that!

PR #356, opened 2023-01-28
Fix attorney address issue #345

austin-bagwell:
AttorneyInput was trying to pass a string into the Select component within AddressInput. The Select component requires an object to be passed into the 'value' prop. The added stateObj will convert any string passed in as the 'state' property of address into an object with the shape expected by Select.

The zipcode issue was a typo: zipCode > zipcode.

### DIFF
--- a/src/components/pages/GenerationPage/GenerationInput/AddressInput.js
+++ b/src/components/pages/GenerationPage/GenerationInput/AddressInput.js
@@ -26,7 +26,8 @@ const TextInput = styled(Input)`
 `;
 
 export default function AddressInput({ address, setAddress, disabled, errors, onClearError }) {
-  const { address1, address2, city, stateAsObject, zipcode } = address;
+  const { address1, address2, city, state, zipcode } = address;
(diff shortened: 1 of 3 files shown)


src/components/pages/GenerationPage/GenerationInput/AddressInput.js
```diff
@@ -26,7 +26,8 @@ const TextInput = styled(Input)`
 `;
 
 export default function AddressInput({ address, setAddress, disabled, errors, onClearError }) {
-  const { address1, address2, city, state, zipCode } = address;
+  const { address1, address2, city, state, zipcode } = address;
+  const stateObj = typeof state === 'string' ? { value: state, label: state } : state;
```
robert-w-gries, line 30: Was the type of the `state` data inconsistent here? If so, I'd like to make it consistent if it's easy enough.

robert-w-gries (commented):
Thanks for the PR! One comment

robert-w-gries (approved):

Issue #358, opened 2023-02-01
Sort offenses chronologically on petition list page


rebecca-draben:
Ali says in Jan 31, 2023 email, "It would be great if some day the offenses were listed chronologically when you click the offense button on the petition list page."

<img width="967" alt="Screen Shot 2023-02-01 at 2 29 20 PM" src="https://user-images.githubusercontent.com/1581141/216144951-1310fd94-79f7-4e33-882c-bbcd11f81ce2.png">
<img width="900" alt="Screen Shot 2023-02-01 at 2 29 35 PM" src="https://user-images.githubusercontent.com/1581141/216144975-6794caca-0496-45a3-bc12-95675c7f713b.png">

PR #360, opened 2023-02-07
Sort offenses chronologically on petition list page

rebecca-draben:
Resolves issue https://github.com/deardurham/dear-petition/issues/358


copelco (approved):
Looks good to me! 

image

The petitions sort offenses oldest first, but I'm not sure they want the same direction here or not!

georgehelman (approved):

PR #366, opened 2023-03-04
Skip duplicate CIPRS records in batch

rebecca-draben:


dear_petition/petition/etl/load.py
```diff
@@ -29,17 +29,24 @@ def import_ciprs_records(files, user, parser_mode, batch_label=""):
     else:
         batch = pm.Batch.objects.create(user=user)
     logger.info(f"Created batch {batch.id}")
+    saved_file_nos = []
     for idx, file_ in enumerate(files):
         logger.info(f"Importing file {file_}")
         batch_file = batch.files.create(file=file_)
+
         for record_data in parse_ciprs_document(batch_file.file, parser_mode):
```
copelco, line 37: Thinking on this, one idea could be to use the DB as the source of truth, so we don't have to pass `saved_file_nos` and `exclude_file_nos` around here and in the other methods:

```suggestion
        for record_data in parse_ciprs_document(batch_file.file, parser_mode):
            file_no = record_data.get("General", {}).get("File No", "")
            if file_no and batch.records.filter(file_no=file_no).exists():
                logger.warning(f"Not saving.. (trimmed)")
                continue
```

It does add a few additional DB queries during load, but the tradeoff could be worth it for the simplicity.

copelco (approved):
Works for me! I added one suggestion, but don't see it as a deal breaker.

dear_petition/petition/etl/load.py
rebecca-draben: Thanks for the suggestion.  That is cleaner.  But I think I'll leave as is for now given that it is a tradeoff with performance, as you said.

PR #371, opened 2023-03-11
Checkmark 3b

georgehelman:


dear_petition/petition/helpers.py
```diff
@@ -24,3 +29,36 @@ def get_285_form_agency_address(agency):
...
+    def calculate_letter_length(letter):
+        if letter in letter_lengths:
+            return letter_lengths[letter]
+        letter_length = get_text_pixel_length(letter)
+        letter_lengths[letter] = letter_length
+        return letter_length
+
+    truncated_string_size = 0
+    idx = 0
+    while True:
```
robert-w-gries, line 55: I think we need to add a check for the length of `text`, otherwise we might hit an infinite loop when the string fits within the `desired_length`.

```suggestion
    while idx < len(text):
```

dear_petition/petition/models.py
```diff
@@ -380,7 +381,7 @@ def get_all_offense_records(self, include_annotations=True, filter_active=False)
 
         return qs
 
-    @property
+    @cached_property
```
robert-w-gries, line 384: Wouldn't this result in a stale document being returned after we call `recalculate_petitions()`?

dear_petition/petition/models.py
```diff
@@ -402,18 +403,13 @@ class PetitionDocument(PrintableModelMixin, models.Model):
         "self", on_delete=models.CASCADE, null=True, related_name="following_document"
     )
     agencies = models.ManyToManyField(Contact, related_name="+")
+    form_type = models.CharField(choices=FORM_TYPES, max_length=255)
+    form_specific_data = models.JSONField(default=dict)
```
robert-w-gries, line 407: I'd like @copelco's input on this, but something about this `form_specific_data` field seems off to me. I think it's because it's so vague that any data could be thrown in there and the keys are arbitrary. In this case, you are using it store `is_checkmark_3b_checked`. But there are multiple 3B checkboxes in this dismissal form. The one you're setting is under the `Petition to Expunge` section and the other one is under `Findings of Fact`. They even have similar language.

I think my overall point is that we'd ideally build out a more robust solution. Maybe subclasses of `PetitionDocument`? Doesn't necessarily need to be done in this PR, but I think it's something we should do after this feature gets merged.

dear_petition/petition/tests/test_helpers.py
robert-w-gries, line 1: Can you add a `get_truncation_point_of_text_by_pixel_size()` test where the text is less than the desired length? For example:

```python
ph.get_truncation_point_of_text_by_pixel_size('Test', 200)
```

dear_petition/petition/types/addendum.py
```diff
@@ -0,0 +1,60 @@
...
+            addendum_3b_document = pm.PetitionDocument.objects.create(
+                petition=petition,
+                previous_document=previous_document,
+                form_type=pc.ADDENDUM_3B,
+            )
+            addendum_3b_document.offense_records.add(*qs)
+            petition.base_document.form_specific_data["is_checkmark_3b_checked"] = True
+            petition.base_document.form_specific_data[
+                "charged_desc_string"
+            ] = "See attachment"
```
robert-w-gries, line 44: The example Jessica provided says `See addendum`

dear_petition/petition/types/addendum.py
```diff
@@ -0,0 +1,60 @@
...
+            petition.base_document.form_specific_data["is_checkmark_3b_checked"] = True
+            truncation_point = ph.get_truncation_point_of_text_by_pixel_size(
+                checkmark_3b_string, ADDENDUM_3B_FIRST_LINE_LIMIT
+            )
+            petition.base_document.form_specific_data[
+                "charged_desc_string"
+            ] = checkmark_3b_string[0:truncation_point]
+            petition.base_document.form_specific_data[
+                "charged_desc_cont_string"
+            ] = checkmark_3b_string[truncation_point:]
```
robert-w-gries, line 55: Well done! 👏

docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
     depends_on:
       - postgres
     volumes:
+      - ./:/code/
```
robert-w-gries, line 15: I think @copelco has tried to avoid pulling in the whole folder like this. In the past, we've added files in the root dir, like frontend config files, by explicitly listing them

dear_petition/petition/types/addendum_3b.py
```diff
@@ -0,0 +1,17 @@
...
+def get_offense_records(petition):
+    active_records = petition.offense_records.filter(petitionoffenserecord__active=True)
+    active_ciprs_records = active_records.values("offense__ciprs_record").distinct()
+    active_record_ids = active_records.values_list("id")
+
+    return OffenseRecord.objects.filter(
+        ~Q(id__in=active_record_ids),
+        offense__ciprs_record__in=active_ciprs_records,
+        action=pc.CONVICTED,
+        offense__verdict=pc.GUILTY,
```
robert-w-gries, line 16: Do you mind adding a comment explaining why these are the queryset conditions? If I'm understanding it properly, it seems that we want all Guilty + Convicted offenses from CIPRS records that have active Dismissed records? I'm not sure about the negated `id__in=active_record_ids` query though

dear_petition/petition/export/tests/test_cr287.py
```diff
@@ -264,3 +267,16 @@ def test_map_offenses__disposition_date(form, offense1, offense_record1):
...
+# Other
+#
+
+
+def test_checkmark_3b(form, offense1, offense_record1):
+    form.map_additional_forms()
+    assert form.data.get("ChargedA")
+    assert not form.data.get("ChargedB")
+    assert not form.data.get("ChargedDesc")
+    assert not form.data.get("ChargedDescCont")
```
robert-w-gries, line 282: Can you add the following testcases?

1. ChargedA and ChargedB are both `False`
2. ChargedA is `True` and ChargedB is `False
3. ChargedB is `True` and `ChargedDesc` is `See addendum`
4. ChargedB is `True` and `ChargedDesc` and `ChargedDescCont` have a record description

robert-w-gries (changes requested):

docker-compose.yml
copelco: Ah yes, it does appear to be double mounting currently. @georgehelman What files were not auto-updating with the existing mounts below?

robert-w-gries (approved):

docker-compose.yml
copelco: If the full tree is mounted, then these files will not be available in the dev container:

https://github.com/deardurham/dear-petition/blob/833fa8fac35e251f5bf7ba8f0d442540ef1d8c95/Dockerfile#L92

PR #372, opened 2023-03-12
Handle duplicate offenses and GTL in Expungable Record Summary

rebecca-draben:


dear_petition/petition/export/documents/expungable_summary.py
```diff
@@ -62,43 +70,76 @@ def generate_context(batch, contact, petitioner_info):
...
-                offense_record.offense.disposition_method
-            ),
-            "disposed_on": offense_record.disposed_on
-        }
-
-        # append offense record to list for the key, but if key doesn't exist yet, create an empty list first
-        table_data.setdefault(key, []).append(offense_record_data)
+        offense_records = list(offense.offense_records.all())
+        if len(offense_records) not in [1, 2]:
+            logger.error(f"Expecting 1 or 2 offense records, but found {len(offense_records)}")
```
robert-w-gries, line 102: Can you change this to `assert(len(offense_records) in [1, 2])`? I think something wrong has happened if we have an offense with no offense records

dear_petition/petition/export/documents/expungable_summary.py
```diff
@@ -62,43 +70,76 @@ def generate_context(batch, contact, petitioner_info):
...
+
+        guilty_to_lesser = False
+        if offense.verdict == VERDICT_GUILTY and len(offense_records) == 2:
+            same_description = offense_records[0].description == offense_records[1].description
+            same_severity = offense_records[0].severity == offense_records[1].severity
+            if same_description and same_severity:
+                # if offense records same, then remove one
+                offense_records.pop()
```
robert-w-gries, line 110: Can you make a helper function like `is_convicted_offense_record()` or similar that encapsulates this logic? If we later decide to modify how we track the convicted offenses, it'd be easier to find in the code and to replace it:

```python
    def is_convicted_offense_record(offense_records):
        if not offense.verdict == VERDICT_GUILTY or not len(offense_records) == 2:
            return False

        same_description = offense_records[0].description == offense_records[1].description
        same_severity = offense_records[0].severity == offense_records[1].severity
        return same_description and same_severity
```

robert-w-gries (changes requested):

dear_petition/petition/constants.py
```diff
@@ -99,14 +99,22 @@
     "PROBATION OTHER": "PO",
     "PROBATION REVOKED": "PR",
     "REMANDED TO DISTRICT COURT": "RM",
-    SUPERSEDING_INDICTMENT_OR_PROCESS: "SI",
+    DISP_SUPERSEDING_INDICTMENT: "SI",
```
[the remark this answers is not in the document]
rebecca-draben: During development, I had made another one of these dispositions a constant (but later removed it because I ended up not needing it).  

When there were two (and there could be two or more again at some future point), it helped with code readability (IMHO) to prefix them with what they are (a disposition). 

Plus, if you don't prefix them with what they are, then sometimes there will be conflicts with other constants that happen to have the same value.  For example I also made a constant for the verdict "GUILTY" `VERDICT_GUILTY = "GUILTY"`.  I could have just called it `GUILTY = "GUILTY"`.  And then another developer might come along and want to use a constant for the plea "GUILTY".  But they really shouldn't be using the same constant for both verdict and plea because one could potentially change without the other.  But if it is prefixed, hopefully the second developer would realize how it is used and make the constant `PLEA_GUILTY = "GUILTY"`.

dear_petition/petition/export/documents/expungable_summary.py
rebecca-draben: Sure, I wasn't sure if it was better to make a best effort at creating the document or throw an error.  I'll change it.

dear_petition/petition/export/documents/expungable_summary.py
rebecca-draben: Good idea.  I'll change.

dear_petition/petition/models.py
```diff
@@ -128,6 +128,32 @@ class Offense(PrintableModelMixin, models.Model):
...
+        """
+        return self.verdict == pc.VERDICT_GUILTY and not self.has_equivalent_offense_records()
+
+    def has_equivalent_offense_records(self):
+        """
+        Return true if the CHARGED AND CONVICTED offense records are equivalent. Raise error if there is an unexpected
+        number of offense records.
+        """
+        offense_records = list(self.offense_records.all())
+        assert (len(offense_records) in [1, 2])
```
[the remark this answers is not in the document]
rebecca-draben: Good catch!

robert-w-gries (approved):

Issue #378, opened 2023-03-16
Use verdict as disposition in Expungable Summary Document if it exists.


rebecca-draben:
If a verdict exists in the CIPRS record, use the verdict as the disposition in the Expungable Summary Document.

Issue #377, opened 2023-03-14
Update Guilty to Lesser logic to change CONVICTED offense to disposition GU in Summary Document


rebecca-draben:
When the verdict is GUILTY and the offense records are different, the disposition on the CONVICTED offense record should be changed to GU.

Issue #376, opened 2023-03-14
Implement Responsible to Lesser logic in Summary Document


rebecca-draben:
If the verdict is RESPONSIBLE, then look to see if the offenses records are the same or different.  If they are the same, drop one offense record from the Summary Document.  

![Image](https://user-images.githubusercontent.com/1581141/225119547-35c6e3bd-4523-4f27-9459-98cc2f9c7125.png)

If they are different, change the disposition in the Summary Document on the CHARGED offense record to be RL and the disposition on the CONVICTED offense record to be RS.

![Image](https://user-images.githubusercontent.com/1581141/225119596-908711ff-61a7-4563-998b-52996e0295d2.png)

PR #384, opened 2023-04-02
Calculate and save offense record disposition and visibility

rebecca-draben:
Disposition will be verdict if it exists. Disposition will be Guilty to Lesser or Responsible to Lesser when applicable.

This addresses 3 issues
https://github.com/deardurham/dear-petition/issues/376
https://github.com/deardurham/dear-petition/issues/377
https://github.com/deardurham/dear-petition/issues/378


dear_petition/petition/models.py
```diff
@@ -150,6 +182,8 @@ class OffenseRecord(PrintableModelMixin, models.Model):
     action = models.CharField(max_length=256)
     severity = models.CharField(max_length=256)
     description = models.CharField(max_length=256)
+    disposition = models.CharField(max_length=256, blank=True)
+    is_visible = models.BooleanField(default=True)
```
robert-w-gries, line 186: I think these fields can be converted into properties as they are deriving data from the offense. It would look like this:

```python
@property
def disposition(self):
    if self.action == CHARGED:
        if offense.is_guilty_to_lesser():
            return VERDICT_GUILTY_TO_LESSER
        elif offense.is_responsible_to_lesser():
            return VERDICT_RESPONSIBLE_TO_LESSER
    return offense.verdict if offense.verdict else offense.disposition_method

@property
def is_visible(self):
    return not (self.action == CHARGED and offense.is_convicted_of_charged())
```

dear_petition/petition/export/documents/expungable_summary.py
```diff
@@ -97,38 +97,40 @@ def __create_tables_data(offenses):
         jurisdiction = JURISDICTION_MAP[offense.ciprs_record.jurisdiction]
         key = (county, jurisdiction)
 
-        offense_records = list(offense.offense_records.all())
-        if offense.is_convicted():
-            # remove one of the offense records because the two offense records would appear identical in the output
-            offense_records.pop()
+        offense_records = list(offense.offense_records.all().filter(is_visible=True))
```
robert-w-gries, line 100: The unfortunate part of model properties is that you can't use them to filter like you can with the database fields. I think it's fine to replace with `[offense_record for offense_record in list(offense.offense_records.all()) if offense_record.is_visible]`.

robert-w-gries (commented):
I left a couple comments regarding the use of properties instead of database fields. After digging into our Offense vs OffenseRecord logic, it's not clear to me what the intention was behind the separation of models.

I'm thinking we might be going about things the wrong way by always passing in offense records. Either we need to apply the various querysets defined in `dear_petition\petition\types\` to extract the proper offense records, or we can store the "proper" offense record data in the `Offense` model. For example, the "proper" offense record data in the case of Guilty to Lesser means setting the `offense.description` to the convicted offense record.

If we treat the `Offense` as the source of truth for exporting to the Petition pdfs or to the Expungable Summary, then we don't need to do all these calculations on the offense records to see if there's a partner offense record with matching data and whether it should be included or not. I think in the ideal case, that's already been done while extracting + transforming the data, and we just iterate over the set of Offenses.

I don't want to delay your changes too long, so I'd like to propose we remove the database fields and proceed with the properties. That way, if we decide to change how the model looks we don't need to worry about data migrations or anything.

And for my own benefit so I can remember this later, I think we should add a `disposition` on the `Offense` model that gets saved after we finish parsing the CIPRS record.

robert-w-gries (approved):

PR #385, opened 2023-04-05
hide Advice Letter button from UI

austin-bagwell:
Temporarily hiding this button at the request of the Legal Aid team

### DIFF
--- a/src/components/pages/GenerationPage/GenerationPage.js
+++ b/src/components/pages/GenerationPage/GenerationPage.js
@@ -191,11 +191,13 @@ function GenerationPage() {
                 {/*
                 Legal team requested this be temporarily removed from UI
                  */}
-                <Button
-                  type="button"
-                  className={cx('w-0', { invisible: true })}
-                  onClick={() => generateAdviceLetter()}
-                >
+                {false && (
+                    <Button
+                      type="button"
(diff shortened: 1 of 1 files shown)


src/components/pages/GenerationPage/GenerationPage.js
```diff
@@ -187,7 +188,14 @@ function GenerationPage() {
...
               <div>
-                <Button type="button" onClick={() => generateAdviceLetter()}>
+                {/*
+                Legal team requested this be temporarily removed from UI
+                 */}
+                <Button
+                  type="button"
+                  className={cx('w-0', { invisible: true })}
+                  onClick={() => generateAdviceLetter()}
+                >
```
robert-w-gries, line 198: `invisible: true` will make this button render in html and hidden in CSS. So theoretically the user can open up dev tools and change the CSS to `display: block;` or whatever. I'd rather just hide it for now by doing the following:

```suggestion
                {false && (
                    <Button
                      type="button"
                      className={cx('w-0', { invisible: true })}
                      onClick={() => generateAdviceLetter()}
                    >
               ) }
```

If our overzealous linter complains about the above, you can just comment it out instead

robert-w-gries (changes requested):

robert-w-gries (approved):

PR #390, opened 2023-04-16
Fix failing remote UT

georgehelman:


robert-w-gries (approved):
Per slack convo, let’s merge this for now and see if the test is still flakey due to OS differences. We can also make a new issue to track whether more work is needed for making our pixel length code more robust

PR #396, opened 2023-04-30
Condense utils and helpers into one file

georgehelman:


dear_petition/petition/tests/test_helpers.py
robert-w-gries, line 1: Should these test be moved to `test_utils.py`?

georgehelman: Yes, good catch. Done.

robert-w-gries (approved):

PR #399, opened 2023-05-08
Remove unused and unaccessed filter_active param

georgehelman:


robert-w-gries:
Does this need to be removed? It is situational, but it could be useful to get a list of the inactive records

georgehelman:
Right now we don't have a use case for it, but I added it back as a param to the OffenseRecordPaginator in case there is one in the future. I do want to make sure it uses the param that is being passed for clarity though.

PR #400, opened 2023-05-13
Rename expungable_records_summary to records_summary

austin-bagwell:
Renamed occurences of expungable_records_summary to records_summary.

There is a variable in records_summary.py that still shows expungable:
`TEMPLATE = "expungable_summary.docx"`
I think this is a named template from the DocxTemplate library. I'm not sure how to access those templates to make that change.

All pytests past and I'm able to download a petition summary from both views on the frontend.


austin-bagwell:
found the template file, thanks @rebecca-draben !

robert-w-gries (approved):

Issue #367, opened 2023-03-07
Enumerate Offenses in Expungable Record Summary Document


rebecca-draben:
Number each offense under each county/jurisdiction.  Jessica mentioned that this would be useful to her because she sometimes loses track of where she is in the offenses list. Example solution shown below.

![Image](https://user-images.githubusercontent.com/1581141/223306128-ca127074-553b-47eb-8b70-06fe17e58f8b.png)

PR #401, opened 2023-05-14
Add index to offense records in Record Summary document

rebecca-draben:
Addresses issue https://github.com/deardurham/dear-petition/issues/367


dear_petition/petition/export/documents/records_summary.py
```diff
@@ -55,11 +55,23 @@ def generate_context(batch, attorney, client):
...
+        # get offense records for county and jurisdiction
+        offense_records = tables_data[table_key]
+
         # sort offense records by file number
         offense_records.sort(key=lambda x: x["file_no"])
-        table = {"county": k[0], "jurisdiction": k[1], "idx": i, "offense_records": offense_records}
+
+        # assign index for each offense record
+        for offense_record_index, offense_record in enumerate(offense_records, start=1):
+            offense_record["idx"] = offense_record_index
```
robert-w-gries, line 67: Is the template doing `idx + 1` as the count? This is a bit of a personal opinion, but I feel like we should keep as much python logic out of the template file as possible because it's a binary file that we can't properly track in git

rebecca-draben: I agree about keeping the logic out of the template.  The template uses the idx value as is. `enumerate(offense_records, start=1)` ensures the index starts at 1 and not 0.

rebecca-draben: Realized I forgot to add checks in the unit tests.  I'll do that.

PR #404, opened 2023-05-21
Fix sortable headers on UsersPage

austin-bagwell:
- Fixed sortable headers on the UsersTable found on UsersPage. 
- Reused existing logic from AgengiesPage to build the URL query parameters and set the sorting order. 
- Adjusted the getUsersQuery in ./service/api.js to append the query parameters to the end of the URL.

Sorting is not yet supported for the Admin column. That will require changes to the back end and will be addressed in a future PR.

### DIFF
--- a/src/components/pages/UsersPage/UsersPage.js
+++ b/src/components/pages/UsersPage/UsersPage.js
@@ -191,13 +191,7 @@ const UsersPage = () => {
               </button>
             </div>
           </TableFlexRow>
-          <UsersTable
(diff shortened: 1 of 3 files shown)


robert-w-gries (approved):

robert-w-gries:
Thanks for fixing this! I think the table code could benefit from re-factoring, but maybe let's do that if we end up adding a management table for something like "Attornies" or "Clients"

PR #414, opened 2023-05-27
Unmount code directory in docker-compose config

georgehelman:


robert-w-gries (approved):
I'm fine with the change. Was this line causing an issue?

georgehelman:
No issue, we just discussed removing it in a previous meeting since we didn't like mounting the whole repo directory.

PR #417, opened 2023-05-29
Add footer to PageBase

austin-bagwell:
Added a basic footer that links to the [Code with Durham](https://www.codefordurham.com/) site. 

Screenshot 2023-05-29 at 12 09 57 PM
Screenshot 2023-05-29 at 12 10 10 PM


robert-w-gries (commented):
Nice, this turned out well! Minor note but is it possible to have it take less vertical space? I'd prefer if it used more horizontal space and less vertical space as the vertical space in our pages is more valuable. Not a big deal if it's not possible, and definitely go for whatever looks best to you

austin-bagwell:
How does this look? I made a horizontal version of the logo which I do think looks better for this. I also added a lot (8rem) of margin top to the footer element. At least on my laptop screen, that means a small scroll down is required to see the footer, assuming table rows or other content are filling the page.  I think all of that margin-top might be excessive, but I'm not really sure.

Screenshot 2023-05-29 at 6 37 03 PM
Screenshot 2023-05-29 at 6 37 12 PM

src/components/pages/PageBase.styled.jsx
```diff
@@ -58,7 +58,16 @@ export const LinkWrapper = styled.div`
 export const PageContentWrapper = styled.section`
   padding: 2rem 4rem;
   flex: 1;
+  min-height: calc(85vh - 50px);
```
robert-w-gries, line 61: Do you mind sharing more about this line was needed?

austin-bagwell: I needed a way to get the footer to stick to the bottom of the page regardless of how much content existed - screenshot below as an example of what the footer looks like without  `min-height: calc(85vh - 50px);` and without content that fills the page.  Setting `min-height: 85vh` on `PageContentWrapper` seemed to solve that by making sure that the content section filled enough space regardless of how much content there actually was.

The `calc()` was intended to account for manually setting the footer height to 50px as in [this example](https://css-tricks.com/couple-takes-sticky-footer/#aa-there-is-calc-reduced-height-wrappers) but after some adjusting I don't think that was necessary. I'll add a commit that removes the min-height calculation as I don't think it's doing anything useful in this context.

Screenshot 2023-05-31 at 7 41 16 AM

robert-w-gries: Can you use flex instead? The wrapper div would be `flex flex-column` with the main content div having `flex-1`. This will result in the main page content taking up as much vertical space as possible with the footer taking as little space as possible.

robert-w-gries: Helpful guide to flex: https://css-tricks.com/snippets/css/a-guide-to-flexbox/#

austin-bagwell: yep, easy change! Thanks for the tip

robert-w-gries (approved):
Sorry for the delay in approving, I wanted to run it locally to see how it looked with the various pages. I think it looks good to merge! Minor nitpick (i.e. optional): I'd prefer the margin between the footer and main content section to be smaller, maybe 1 or 2 rem. But feel free to merge in whatever you end up liking

PR #428, opened 2023-06-19
Add delete button to app

georgehelman:


src/features/ExistingPetitions.jsx
```diff
@@ -112,6 +112,24 @@ export const ExistingPetitions = () => {
...
+                        responseType: 'arraybuffer',
+                        method: 'delete',
+                      }).then(() => {
+                        setRows(
+                          rows.filter((row) => {
+                            row.pk !== batch.pk;
+                          }),
+                        );
+                      });
+                    }}
```
robert-w-gries, line 129: I recommend using RTK Query here instead. You will not need to manually track the rows state and it will gracefully handle the re-fetching of updated/deleted/added batches using [automated re-fetching](https://redux-toolkit.js.org/rtk-query/usage/automated-refetching).

There's a few steps needed to use RTK Query:

1. We should add an RTK Query endpoint called `deleteBatch`:
```jsx
    deleteBatch: builder.mutation({
      query: ({ id }) => ({ url: `batch/${id}`, method: 'delete' }),
      invalidatesTags: ['Batch'],
    }),
```

2. Call the newly generated `useDeleteBatchMutation` hook at the top of our `ExistingPetitions` component:
```jsx
  const { data } = useGetUserBatchesQuery({ user: user.pk });
  const [triggerDeleteBatch] = useDeleteBatchMutation();
```

3. Call the `triggerDeleteBatch` function in the `onClick` handler of your new button. This will automatically trigger a re-fetch of the user's batch list so we do not need to update the row state manually.
```jsx
                 <Button
                    title="Delete the batch."
                    onClick={() => triggerDeleteBatch({ id: batch.pk })}
                  >
                    Delete
                  </Button>
```

robert-w-gries (changes requested):

robert-w-gries (approved):

PR #437, opened 2023-09-26
Frontend Tests, hooks and utils

austin-bagwell:
Using Vitest as testing framework, this sets up some basic tests for custom hooks and utility functions. More tests to come!


robert-w-gries (changes requested):

src/__tests__/util/isChrome.test.js
```diff
@@ -0,0 +1,63 @@
...
+    const isChrome =
+      !isIOSChrome &&
+      isChromium !== null &&
+      typeof isChromium !== 'undefined' &&
+      vendorName === 'Google Inc.' &&
+      isOpera === false &&
+      isIEedge === false;
+
+    return isChrome;
+  };
```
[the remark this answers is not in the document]
austin-bagwell: Having some issues with this one. Each of the two assertions I'm making seem to be correct on their own, but will only pass testing if they are the first in the series to be tested. I suspect it has to do with importing `isChrome.js` but I'm not totally sure how to begin troubleshooting that. 

Any idea where I should be looking at to get around this?

I'm only testing two assertions in this example
- browser is Chrome and isChrome = true
- browser is Mozilla and isChrome = false

Setting window values with `vi.stubGlobal` is working as expected. Mocking those values and dynamically importing `isChrome` works as expected on the first test, but the second test always fails, regardless of the order that the tests are in. Put another way, each of the tests will pass, but only if they are the first test to be ran. 

Below is an example where I ran `npm run test`, switched the order of the tests and ran `npm run test` again.
// browser=chrome first, then browser=mozilla
```
 ❯ src/__tests__/util/isChrome.test.js (2)
   ❯ isChrome test (2)
     ✓ browser is Chrome, isChrome=true
     × browser is Mozilla, isChrome=false
```
// browser=mozilla first, then browser=chrome
```
❯ src/__tests__/util/isChrome.test.js (2)
   ❯ isChrome test (2)
     ✓ browser is Mozilla, isChrome=false
     × browser is Chrome, isChrome=true
```

I am resetting my vitest environment before each test. ```
  beforeEach(() => {
    vi.clearAllMocks();
    vi.resetAllMocks();
  });

  afterEach(() => {
    vi.stubGlobal('chrome', null);
    vi.stubGlobal('navigator', {
      userAgent: '',
      vendor: '',
    });
```
The import of `isChrome.js` happens after I've reset the necessary `vi.stubGlobal` values.
```
vi.stubGlobal('chrome', true);
    vi.stubGlobal('navigator', {
      userAgent: `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36`,
      vendor: 'Google Inc.',
    });

const result = (await import('../../util/isChrome.js')).default;
```


@robert-w-gries

File: src/__tests__/hooks/useOnClickOutside.test.js [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+
+  it('should not call the handler after hook is unmounted', () => {
+    const ref = { current: document.createElement('div') };
+    const handler = vi.fn();
+
+    const { unmount } = renderHook(() => useOnClickOutside(ref, handler));
+
+    act(() => {
+      const insideElement = document.createElement('div');
+      ref.current.appendChild(insideElement);
```
robert-w-gries, line 54: I think we should use the `outsideElement` code here instead as the handler would only be called if the click was on an element outside the one hosting the hook.

src/__tests__/hooks/useWindowSize.test.js
```diff
@@ -0,0 +1,36 @@
...
+    vi.stubGlobal('innerWidth', 1920);
+    vi.stubGlobal('innerHeight', 1080);
+
+    const { result } = renderHook(() => useWindowSize());
+
+    expect(result.current.width).toBe(1920);
+    expect(result.current.height).toBe(1080);
+
+    vi.unstubAllGlobals();
+  });
```
robert-w-gries, line 16: FYI you can use `beforeEach` and `afterEach` to avoid having to copy + paste the setup and teardown code:

```
describe('useWindowSize', () => {
   beforeEach(() => {
       // vi.stubGlobal
   });
   afterEach(() => {
       // vi.unstubAllGlobals
   });
});
```

src/__tests__/hooks/useOnClickOutside.test.js
austin-bagwell: Pushed through new changes, but I'm not 100% sure that my test is doing the right thing.

I changed this so that the event is dispatched to outerElement after unmounting the ref element hosting the hook, but for some reason it only works if I don't use `act()`.  Keeping the both `unmount()` and `outsideElement.dispatchEvent(event)` inside the `act()` scope caused the test to fail, removing act results in a failed test with the same assertion.

I read the [docs](https://testing-library.com/docs/preact-testing-library/api/#act) and [this](https://github.com/threepointone/react-act-examples/blob/master/sync.md) related post, but I'm still not 100% sure what is going on.

Am I actually testing the correct behavior this way?

@robert-w-gries

austin-bagwell: AB - use RTL debug to check that all DOM elements are present. Can also get more specific with assertions to double check that correct behavior is being tested

src/__tests__/hooks/useOnClickOutside.test.js
```diff
@@ -56,9 +52,14 @@ describe('useOnClickOutside', () => {
...
+
     outsideElement.dispatchEvent(event);
+    expect(handler).toHaveBeenCalledTimes(1);
 
-    expect(handler).not.toHaveBeenCalled();
+    unmount();
+
+    outsideElement.dispatchEvent(event);
+    expect(handler).toHaveBeenCalledTimes(1);
+    expect(handler).not.toHaveBeenCalledTimes(2);
```
robert-w-gries, line 63: nitpick: if you've already checked it was called once, the second assert is redundant

robert-w-gries (approved):

PR #441, opened 2023-09-30
285 documents should have same file number as base document, add et al

georgehelman:


georgehelman:
https://github.com/deardurham/dear-petition/issues/442

dear_petition/petition/export/forms.py
```diff
@@ -112,9 +72,7 @@ def map_header(self):
             self.data["Superior"] = Checkbox("")
 
     def map_file_no(self):
-        first_record = self.get_first_record()
-        if first_record:
-            self.data["ConsJdgmntFileNum"] = first_record.offense.ciprs_record.file_no
+        self.data["ConsJdgmntFileNum"] = self.extra["file_no"]
```
robert-w-gries, line 75: Would the following work and be simpler? I renamed `add_additional_params_to_forms` to `get_file_number_from_documents` to make the function more specific about what it's doing.

```py
        self.data["ConsJdgmntFileNum"] = get_file_number_from_documents(self.petition.documents)
```

georgehelman: That would work, but it would be re-running the query once for every form, versus hoisting the query above the forms and running it once. Also, I named it more generally because I thought we might want to do this type of thing again in the future, so I wanted to create a pattern for adding data that is required on all forms.

robert-w-gries: I get the reasoning for splitting it out. My preference is still for keeping in the same file/loop but I'm okay with merging as-is

robert-w-gries (approved):

Issue #464, opened 2024-02-07
When some charges dismissed in District and some resolved in Superior, dismissed charges appearing on Superior petition


rebecca-draben:
From Emily's Oct 19, 2023 email: For a record where the client has some charges dismissed in District Court and some charges resolved in Superior Court, the tool puts all the dismissed charges on a Superior Court petition. Instead, the ones that are dismissed in District Court should go on a District Court petition.

(see email for specific examples)

copelco:
We think this is resolved. @georgehelman to investigate

copelco:
From today's meeting: Fix was implemented for adult misdemeanors in #462, need to add to dismissed, guilty, and underage convictions.

georgehelman:
PR here: https://github.com/deardurham/dear-petition/pull/471

PR #471, opened 2024-03-23
Previously, we went by the jurisction of the containing CIPRS record …

georgehelman:
…when determining the jurisdiction of the offenses. However, it is possible for a CIPRS record to contain both district and superior court offenses in cases where the offense has a superseding indictment. We should always go by which section the offense is in (District Court Offense Information vs Superior Court Offense Information) when determining the offense's jurisdiction.


georgehelman:
https://github.com/deardurham/dear-petition/issues/464

dulshen (approved):

PR #476, opened 2024-05-04
Pin lxml package so Heroku build hopefully succeeds

georgehelman:


copelco (approved):
Seems worth the try to me!

PR #479, opened 2024-05-20
Make extended timeout specific to /batch/{pk}/ GET requests

rebecca-draben:
And set default timeout back to previous value

### DIFF
--- a/src/service/api.js
+++ b/src/service/api.js
@@ -90,7 +90,7 @@ export const api = createApi({
       },
     }),
     getBatch: builder.query({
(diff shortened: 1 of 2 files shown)


georgehelman (approved):

robert-w-gries:
Which endpoint(s) do you want to keep the increased timeout for? You can add the timeout option here in the endpoint definition file: https://github.com/deardurham/dear-petition/blob/master/src/service/api.js#L77

rebecca-draben:
Ah, yes @robert-w-gries that does look better.

Issue #489, opened 2024-07-28
Change generated petition titles


georgehelman:
LANC has adopted a new file naming convention (YEAR-MONTH-DAY DOCUMENT TYPE Client Name). We have a request to change the file names generated by the robot.

Currently, petition file names: CLIENTNAME FORMNUMBER DISTRICT/SUPERIOR COUNTY

We request instead: DATEGENERATED COUNTY DISTRICT/SUPERIOR FORM-STATUTE CLIENTNAME

We’re asking to change from form number to statute number, because the forms may change but the statutes are less likely to change.

Form Number | Statute | Provision | Naming

285 | n/a | Attachment | n/a
287 | 146(a) | Charges Dismissed | 146
288 | 146(a2) | Not Guilty or Not Responsible | 146NG
293 | 145.8A | Raise the Age | 145.8A
297 | 145.5F (F for felony) | Nonviolent Felony(ies) | 145.5F
298 | 145.5M (M for misdemeanor) | Nonviolent Misdemeanor(s) | 145.5M

Example
Current Orville Shepherd AOC-CR-287 D Lee
New 2024-05-15 Lee SC 287 Orville Shepherd

georgehelman:
https://github.com/deardurham/dear-petition/pull/490

PR #490, opened 2024-07-28
Change petition titles to new AOC format

georgehelman:
https://github.com/deardurham/dear-petition/issues/489


dear_petition/petition/tests/test_utils.py
```diff
@@ -1,11 +1,14 @@
 import pytz
+import pytest
 from django.utils.timezone import make_aware, utc
 from datetime import datetime, date
 
 from dear_petition.petition import utils as pu
 
 from ..constants import DATETIME_FORMAT
 
+pytestmark = pytest.mark.django_db
```
robert-w-gries, line 10: What does this do?

georgehelman: This tells pytest that we need the test database for this test suite so we can get access to test fixtures like `petition`, which I use in my new test.

robert-w-gries (approved):

PR #503, opened 2024-09-29
Record import export

georgehelman:


src/features/NewPetitionFromRecordSpreadsheet.jsx
```diff
@@ -0,0 +1,183 @@
...
+import { useCreateBatchFromRecordSpreadsheetMutation } from '../service/api';
+import { Spinner } from '../components/elements/Spinner';
+import { POSITIVE } from '../components/elements/Button/Button';
+
+const ALLOWED_MIME_TYPES = [
+  'application/vnd.ms-excel',
+  'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+];
+const MAX_FILES = 1;
+const MAX_FILE_SIZE = 30000;
```
robert-w-gries, line 17: Do we really want this max file size? I can't tell if it's in bytes, KB or MB, so it may be pretty small.

dear_petition/petition/api/viewsets.py
```diff
@@ -407,6 +427,26 @@ def assign_client_to_batch(self, request, pk):
...
         return Response({"batch_id": batch.pk})
+    
+    @action(
+        detail=False,
+        methods=[
+            "post",
+        ],
+    )
+    def import_spreadsheet(self, request):
+        files = request.data.getlist("files")
```
rebecca-draben, line 438: Consider try/except block to gracefully handle errors.  Also do you need to make the transaction atomic?

dear_petition/petition/resources.py
```diff
@@ -22,10 +45,86 @@ def parse_agency_full_address(full_address: str) -> typing.Tuple[str, str, str,
...
+            self.ws = self.wb.worksheets[sheet_index]
+        else:
+            raise ValueError("You must provide either a sheet name or a sheet index.")
+
+    def create_new_worksheet(self, title=None):
+        self.wb.create_sheet(title=title)
+        self.change_worksheet(sheet_name=title)
+
+    def append(self, values, fields, is_header=False, num_indent=0, bold=False, color=None,):
+        assert len(values) == len(fields), "The values and fields passed to append should match."
```
rebecca-draben, line 75: Better to raise error.  I believe asserts can be disabled through configuration.

dear_petition/petition/resources.py
```diff
@@ -73,4 +172,289 @@ def skip_row(self, instance, original, row, import_validation_errors=None):
...
+        self.offense_resource = OffenseResource()
+        self.offense_headers = self.offense_resource.get_export_headers()
+        self.first_offense_header = self.offense_headers[0]
+        self.offense_record_resource = OffenseRecordResource()
+        self.offense_record_headers = self.offense_record_resource.get_export_headers()
+        self.first_offense_record_header = self.offense_record_headers[0]
+
+    def export(self, batch_object=None, *args, **kwargs):
+        # Initialize a new dataset with headers
+        assert isinstance(batch_object,models.Batch), "Queryset must be a Batch object"
```
rebecca-draben, line 369: Better to raise error.  I believe asserts can be disabled through configuration.

dear_petition/petition/api/viewsets.py
georgehelman: I added an atomic block to the import_data method of the resource

src/features/NewPetitionFromRecordSpreadsheet.jsx
georgehelman: I got rid of it.

src/features/ExistingPetitions.jsx
```diff
@@ -227,7 +227,25 @@ export const ExistingPetitions = () => {
                       });
                     }}
                   >
-                    Records Summary
+                    Summary
+                  </Button>
+                  <Button
+                    disabled={!!batch?.generate_summary_errors?.batch}
+                    title={batch?.generate_summary_errors?.batch?.join(' ') ?? ''}
```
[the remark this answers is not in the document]
georgehelman: Okay, I got rid of it.

dear_petition/petition/api/viewsets.py
rebecca-draben: So does it handle it gracefully if user imports spreadsheet with unexpected format?  Or is try/except needed?

src/components/pages/GenerationPage/GenerationPage.jsx
```diff
@@ -186,7 +186,6 @@ function GenerationPage() {
             </Button>
             <Button
               onClick={() => generateRecordsSpreadsheet()}
-              disabled={!!data?.generate_summary_errors?.batch}
               title={data?.generate_summary_errors?.batch?.join(' ') ?? ''}
```
robert-w-gries, line 189: This should be removed as well. Otherwise, it will show errors for the account summary generation while hovering over the button

robert-w-gries (approved):

Issue #534, opened 2025-03-10
dob warning message shows up initially when creating a client in certain case


dulshen:
dob warning message shows up initially when creating a client in certain case where create client option is clicked by user but the batch already has an associated client

dulshen:
this was resolved by PR #538

Issue #529, opened 2025-02-19
Look into extensions to react hook forms that allow non-blocking warning messages


dulshen:
React hook forms only seems to allow for a single type of error message, and this is connected to validation so it prevents form submission. In the case of date of birth for a petitioner, we need to be able to have non-blocking warning messages. This is currently implemented using react state directly, but we should look into an extension to react hook form that can handle this instead, as it would be a cleaner solution.

dulshen:
I looked into this some and found possibilities for using resolvers with an external validation library, however none of these really seemed to offer the possibility of soft validation (non-blocking warnings) as far as I could tell.

I worked on a custom hook to encapsulate the warnings logic, as an alternative, which I think really cleans up the code relating to warnings. I'll plan to PR this soon to see if this seems like a good solution.

PR #536, opened 2025-03-14
Refactoring dob warnings to encapsulate form warning logic in a custom hook

dulshen:
Moved logic for dob warnings to a custom hook and made it more general so it could be reused for other fields more easily. I think this greatly simplifies the code needed to use warnings on the Component side, so might be an acceptable solution for issue #529 since there did not appear to be existing extensions to react-hook-form that do this.

Refactoring in this way also resolves issue #534 since the issue no longer occurs when using a custom hook implementation.


robert-w-gries:
Accessibility: https://webaim.org/resources/contrastchecker/

RGB to Hex: https://www.google.com/search?client=firefox-b-1-d&q=rgb+to+hex

Hex value I used during call: #B06203

dulshen:
closing this PR as it was decided that custom hook added too much complexity

PR #542, opened 2025-05-17
Make disposition data come from the DispositionEvents endpoint that p…

georgehelman:
…opulates the RegisterOfActions service in Portal. This returns easy to work with JSON data. This is good as the web scraping did not work with multiple dispositions


bookmarklet/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@code-with-durham/bookmarklet",
-  "version": "0.1.0",
+  "version": "0.2.3",
```
robert-w-gries, line 3: How did you pick this specific version?

dear_petition/portal/etl/extract.py
```diff
@@ -4,14 +4,14 @@
 from .parsers import case_summary, dispositions, case_info, party_info
 
 
-def extract_portal_record(source):
+def extract_portal_record(source, record_id: str):
```
robert-w-gries, line 7: Should record_id be optional for backwards compatibility?

dear_petition/portal/views.py
```diff
@@ -18,10 +18,11 @@ def bookmarklet_handler(request):
     if not form.is_valid():
         logger.warning(f"Bookmarklet validation failed: {form.errors.as_json()}")
         return HttpResponse(status=400)
-
+    breakpoint()
```
robert-w-gries, line 21: This seems unintentional

bookmarklet/package.json
georgehelman: I went through several iterations of committing versions that didn't work. I think the version 0.2.3 evokes a sense of progression and constant improvement.

dear_petition/portal/etl/extract.py
georgehelman: I can do that.

dear_petition/portal/views.py
georgehelman: I will remove this

bookmarklet/package.json
robert-w-gries: The versioning of packages has semantic meaning as npm packages generally follow semver. It’s fine for now, but going forward I recommend doing a minor version bump for a change like this since it’s backwards compatible.

PR #558, opened 2025-10-28
add user feedback widget from sentry to front end

dulshen:
added sentry to the front end, and initialized it so that only the user feedback widget is used


dulshen:
Setting up the DSN in sentry/react will require adding a config variable to staging/production environments (VITE_SENTRY_DSN). Have not tested actually sending feedback yet, wanted to discuss whether this should use a separate DSN for front-end (or possibly 2 separate DSNs one for staging one for production).

georgehelman (approved):

georgehelman (approved):

